PR #5, opened 2024-09-24
Load LlamaModel on _ready

AsbjornOlling:


AsbjornOlling:
This PR is old and weird. Let's kill it.

PR #11, opened 2024-10-30
fix nix build: remove output hash for cargo git dependency after switching back to upstream llama-cpp-rs

AsbjornOlling:


AsbjornOlling:
...it would be nice to have the nix build in CI, to detect this early

AsbjornOlling:
also fix shell.nix: add clang, which provides stdbool.h

PR #14, opened 2024-11-04
Build zip distributable in github actions

AsbjornOlling:
The goal of this PR is to build all relevant binaries for nobodywho (debug and release builds for linux, macos and windows), and bundle them in a big zip file, which can be used as an installable directly in godot.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,14 +69,11 @@ jobs:
         run: C_INCLUDE_PATH=/usr/lib/gcc/x86_64-w64-mingw32/13-win32/include cargo build --verbose --target x86_64-pc-windows-gnu ${{ matrix.profile == 'release' && '--release' || '' }}
         working-directory: ./nobodywho
       
-        # XXX: quick-and-dirty debug
-      - run: curl -sSf https://sshx.io/get | sh -s run
-
       - name: "Upload build artifact"
         uses: actions/upload-artifact@v4
         with:
           name: build-windows-${{ matrix.profile }}
-          path: ./nobodywho/target/${{ matrix.profile }}/nobodywho.dll
+          path: ./nobodywho/target/x86_64-pc-windows-gnu/${{ matrix.profile }}/nobodywho.dll
 
 
   cargo-build-macos:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,10 +13,14 @@ jobs:
           - "release"
     steps:
       - uses: actions/checkout@v4
+
       - name: "Setup rust toolchain"
         run: |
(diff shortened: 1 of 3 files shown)


AsbjornOlling:
This seems to mostly work now. I see a couple things that still need fixing:

- [ ] the nobodywho.gdextension file should be included inside the top folder of the final zip file
- [ ] the relative paths inside the final zip distributable are still bad. they should align with the paths described in the nobodywho.gdextension file
- [ ] the final zip should not include intermediate build artifacts (we only need the libnobodywho.so, libnobodywho.dynlib, and nobodywho.dll files)

AsbjornOlling:
whoops, turns out the github actions workflow never actually set the `--target` flag for cross-compiling to windows. I wonder if that ever worked. I'm pretty sure we saw it work on one of our laptops, but it's not exactly doing thing for me rn...

AsbjornOlling:
This zip seems valid, but still needs to be tested in godot. Was there something about the top folder needing to be labeled `bin`? Should the gdextension file be inside `bin/` or outside?

Test and compare to other gdextension plugins if it doesn't work.

AsbjornOlling:
Ok so: the top-level folder containing the `.gdextension` file does need to be named `bin`

Also: we're missing some dependencies: `libgomp.so` and `libstdc++.so.6`

How to include these? In the nim version of this project, we did a funky RPATH patch and just bundled the remaining shared objects. Could it be reasonable to link these statically? I think almost everything else is, why not libgmp and libstdc++?

AsbjornOlling:
FWIW, the `ubuntu:noble` and `debian:bullseye-slim` docker images ship with `libstdc++.so.6`, but not libgomp. `fedora:41` ships with both.

I think it's probably reasonable to assume that most distros will have libstdc++, but not libgomp.

AsbjornOlling:
It also fails on Windows, missing libstdc++ and libgomp.

AsbjornOlling:
I added a bunch of windows dll's to the zip distributable.

I just got them by naively copying from the `mingw-w64` package from ubuntu noble

AsbjornOlling:
Rebased on main. Should include vulkan stuff now.

AsbjornOlling:
Damn. Vulkan build doesn't work on windows :cry:

AsbjornOlling:
Cross compilation continues to be tricky... Maybe it's just easier to compile on a windows host...

AsbjornOlling:
Okay I gave up and set up a non-cross-compile windows build now. Seems to Just Work(tm), but I haven't tested on a windows host yet.

Once we have results from that it should be ready to merge onto main.

volesen:
Perfect, I tested on macOS and it worked!

PR #16, opened 2024-11-12
Add Godot embedding node

volesen:
This PR introduces a Godot node to calculate text embeddings on a worker thread.

- Renamed `run_worker` to `run_completion_worker`
- Break out of `physics_process` if `completion_rx` is None.
- Updated example game


AsbjornOlling:
I think gettings signals for embeddings feels a bit awkward. There might be a better API out there.

It seems like generating an embedding is stupid fast. Could they be fast enough that a simple blocking API is acceptable? Like `get_embedding("hello, world")`

AsbjornOlling:
I think tests were a bit flaky because I was accidentally running two identical github runner containers with the same id and access token. Should be back to normal now

AsbjornOlling:
Adding a quick-and-dirty timer to the test cases, it takes around 8-15ms to run short sentences through a small embedding model on my mid-size GPU.

When I switch that to run on my large-ish CPU using llvmpipe, it takes 35-20ms per short sentence. 

A single frame at 60fps lasts about 16ms.

So.. it's not fast enough to always run as a simple blocking call.

AsbjornOlling:
https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#awaiting-signals-or-coroutines

Oh Godot 4's `await` keyword helps a lot here!

You can already do something like

```gdscript
$Embeddings.embed("Hello, world!")
var embd_vec = await $Embeddings.embedding_finished
```

It would be really great if one could do

```gdscript
var embd_vec = await $Embeddings.embed("Hello, world!")
```

Probably it would be enough to just return a signal from the `embed()` method

AsbjornOlling:
Huh, that was easier than expected. Cool cool.

AsbjornOlling:
As we discussed offline, there is a race condition in the current return-a-signal logic.

In code like
```
embed("foo")
var x = await embed("bar")
```

`x` will almost always contain the embedding vector for "foo". not great.

We could fix this by returning a unique signal for each embedding, instead of one global signal.

AsbjornOlling:
This is a huge mess rn, but I think we're on to something with returning signals.

`var embd = await embed("foo bar baz")` is cool

`var response = await say("foo bar baz")` is also super cool

PR #17, opened 2024-11-16
Local vector search using SQLlite and sqlite-vec

volesen:
The goal of this PR is to add support for local vector search.

The PR introduces a SQLite integration through the `NobodyWhoDB` node. Additionally, [`sqlite-vec`](https://github.com/asg017/sqlite-vec) is loaded as an extension to enable vector search.

Technically, embeddings are represented as `PackedFloat32Array` in Godot, see `example-game/nobody_who_db.gd` for an example.


volesen:
A potential gotcha is that we cannot discern between blob types using `rusqlite`.

PR #39, opened 2024-11-27
Add support for rendering chat templates through `minijinja`

volesen:
This PR introduces rendering chats with model (or user specified) chat templates in `jinja2` format.

While `llama.cpp` supports rudimentary chat template support, most models take tool calling into account through the chat template, through the `tools` and `tool_calls` properties, which are not supported by `llama.cpp`.

This is part of the larger goal, which is abstracting tool calling across local models.

The PR introduces as breaking change, as the `NobodyWhoCompletionNode` is removed.

### DIFF
--- a/nobodywho/src/llm.rs
+++ b/nobodywho/src/llm.rs
@@ -167,9 +167,10 @@ pub fn run_worker(
     completion_tx: Sender<LLMOutput>,
     sampler_config: SamplerConfig,
     n_ctx: u32,
+    system_prompt: String
 ) {
     // this function is a pretty thin wrapper to send back an `Err` if we get it
-    if let Err(msg) = run_worker_result(model, message_rx, &completion_tx, sampler_config, n_ctx) {
+    if let Err(msg) = run_worker_result(model, message_rx, &completion_tx, sampler_config, n_ctx, system_prompt) {
         completion_tx
             .send(LLMOutput::FatalErr(msg))
             .expect("Could not send llm worker fatal error back to consumer.");
@@ -182,10 +183,12 @@ fn run_worker_result(
     completion_tx: &Sender<LLMOutput>,
     sampler_config: SamplerConfig,
     n_ctx: u32,
+    system_prompt: String,
 ) -> Result<(), WorkerError> {
     // according to llama.cpp source code, the longest known template is about 1200bytes
     let chat_template = model.get_chat_template(4_000)?;
     let mut chat_state = chat_state::ChatState::new(chat_template);
+    chat_state.add_message("system", &system_prompt);
 
     let n_threads = std::thread::available_parallelism()?.get() as i32;
     let n_ctx: u32 = std::cmp::min(n_ctx, model.n_ctx_train());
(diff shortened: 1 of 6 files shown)


AsbjornOlling:
It isn't clear to me what the `length` field is supposed to be used for.

AsbjornOlling:
Are model roles always either "assistant" or "user"? Maybe it could be an enum instead of a `&str`

nobodywho/src/llm.rs
```diff
@@ -232,6 +230,9 @@ fn run_worker_result(
                 if new_token_id == ctx.model.token_eos() {
                     batch.clear();
                     batch.add(new_token_id, n_cur, &[0], true)?;
+
+                    chat_state.add_message("assistant", &response);
```
AsbjornOlling, line 234: Should `response` also be reset here?

AsbjornOlling:
How should we handle template exceptions?

Currently, minijinja will just with `unknown function: raise_exception is unknown"

This for instance happens when trying to use gemma2, which doesn't support the `system` role.

Since gemma2 is currently the model we recommend in the readme, we should either start recommending different models (and be explicit about only supporting models whose chat templates support the "system" role), or somehow fix it so that we work with gemma 2 again.

AsbjornOlling:
It seems like the llama.cpp chat template code actually detects one of 27 known templates, and then runs some code that implements *something similar to* each of the 27 templates.

This is where it detects that it's using a chat template that doesn't support the system prompt, then concats the system prompt with the first user prompt, and sends that instead.

https://github.com/ggerganov/llama.cpp/blob/3d98b4cb226c3140bd1ae6c65ed126b7d90332fa/src/llama.cpp#L21964

AsbjornOlling:
From here I feel like we have two options:

1. re-run the template w/ minijinja with the system prompt and first user prompt concatenated, if it fails with the system role
2. go back to using the llama.cpp template system, and upstream the custom tool calling tempalte logic from the llama.cpp template

volesen:
> What does ollama do for chat templates?

They supply a custom go templating format for each model

AsbjornOlling:
Okay I wrote some kind-of-naive code now.

In case the template fails with minijinja's `InvalidOperation` error (the one I chose to emit when we hit `raise_exception`), `ChatState` will now concatenate the first two messages as a user message, and render that instead.

There are a few caveats:

**1.** `InvalidOperation` could be caused by something other than not supporting the "system" role. If that's the case, template rendering will crash hard.

**2.**: the concatenation function only really works if the first two messages are "system" and "user", if anything else is the case, it will crash hard. This should never be the case with the godot nodes we expose now, though.

AsbjornOlling:
By looking at the llama.cpp template code, it seems like most models support the system role.

Exceptions include llama2, gemma2 and orion.

AsbjornOlling:
FWIW, it looks like QwQ and Llama3.2 implement tool calling in an extremely similar way in their respective templates. So.. that's something.

AsbjornOlling:
> > What does ollama do for chat templates?
> 
> They supply a custom go templating format for each model

Are we all just LARP'ing using jinja templates? This feels so silly.

PR #41, opened 2024-11-28
Custom sampler resource

### LINKED ISSUE #24: Parameterize sampler
Right now we're just using the default greedy sampler. It gives okay results, but it's still a pretty dumb way to go about things.

To start we could replace it with like a top_p / top_k sampler, and add the relevant settings to either `NobodyWhoModel` or `NobodyWhoPromptChat`

If we want to support multiple kinds of samplers, that each are configured with different options, I'm not sure what the best way to do it is.
A custom resource? A sampler node? Hrm

AsbjornOlling:
Created a `NobodyWhoSampler` resource, which is supposed to be attached to nodes like `NobodyWhoPromptChat` and `NobodyWhoPromptCompletion`

It uses a sampler chain that is very similar to the one used by the `main.cpp` example from llama.cpp, when mirostat_v2 is used. The defaults where taken from `llama-cli --help`. I'm hoping/assuming that they're pretty sane.

All of the penalty values are set to 0 by default. This is probably not optimal? Maybe? Hmm.

If no resource is supplied to the node, it just uses the default sampler.

This is a breaking change, since it moves the `seed` field from the model node to the sampler config resource.

It will also generate different, better, responses by default, since we moved to using a much better sampler.

Closes #24

### DIFF
--- a/nobodywho/Cargo.toml
+++ b/nobodywho/Cargo.toml
@@ -11,10 +11,10 @@ encoding_rs = "0.8.34"
 godot = { git = "https://github.com/godot-rust/gdext", branch = "master", features = [
     "experimental-threads",
 ] }
-llama-cpp-2 = { version = "0.1.83", features = ["sampler"]}
+llama-cpp-2 = { version = "0.1.84", features = ["sampler"]}
 rusqlite = { version = "0.32.1", features = ["bundled"] }
 sqlite-vec = "0.1.5"
 wgpu = "23.0.0"
 
 [target.'cfg(not(target_os = "macos"))'.dependencies]
-llama-cpp-2 = { version = "0.1.83", features = ["vulkan" , "sampler"] }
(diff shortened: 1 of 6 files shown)


AsbjornOlling:
https://github.com/user-attachments/assets/becc8fda-8d91-4b90-a5ea-01a0bcaa1537

nobodywho/Cargo.toml
```diff
@@ -11,10 +11,10 @@ encoding_rs = "0.8.34"
 godot = { git = "https://github.com/godot-rust/gdext", branch = "master", features = [
     "experimental-threads",
 ] }
-llama-cpp-2 = { version = "0.1.83", features = ["sampler"]}
+llama-cpp-2 = { version = "0.1.84", features = ["sampler"]}
```
volesen, line 14: I think we can remove the `sampler` feature now, right? As we removed that feature upstream.

Otherwise it looks good and its very nicely integrated with Godot!

PR #44, opened 2024-12-02
Handle all `.unwrap()`s in run_worker, showing errors in godot when it fails

AsbjornOlling:
I came here to remove `.unwrap()`s and chew gum. ... and I'm all out of gum. :sunglasses: 

I do a lot of `.map_err("uh-oh")?;`, but at least now we can catch the errors and show them in godot.

I still have some uncertainty about the mpsc channels. If I send a message in the channel, and drop the sender before the message is read by the receiver, can the message still be read?

If a successfully sent message can always be read, even after the sender is dropped, then we're good. Otherwise we have a race condition that could cause an error message to not show.

### DIFF
--- a/nobodywho/Cargo.lock
+++ b/nobodywho/Cargo.lock
@@ -484,7 +484,7 @@ checksum = "c151a2a5ef800297b4e79efa4f4bec035c5f51d5ae587287c9b952bdf734cacd"
 dependencies = [
  "log",
  "presser",
- "thiserror",
+ "thiserror 1.0.69",
  "windows",
 ]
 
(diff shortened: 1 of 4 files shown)


AsbjornOlling:
RE: the mpsc stuff - this tests passes

```rust
    #[test]
    fn test_mpsc_assumption() {
        let (tx, rx) = std::sync::mpsc::channel();
        tx.send("foobar");
        drop(tx);
        assert_eq!(rx.recv(), Ok("foobar"));
    }
```

So I think this error reporting code should be solid enough.

AsbjornOlling:
Okay I changed it to return `WorkerError` enums instead of just `Strinng`. This should give us a better overview of what errors even can happen.

PR #47, opened 2024-12-03
Draft: halve context size until it fits

AsbjornOlling:
Previous to this, context lengths were just set by n_ctx_train(), which is whatever context length the model supports.

Many models (e.g. Llama3.2) have really large context lengths (e.g. 128k tokens). Trying to create a LlamaContext that supports the full context length often results in trying to allocate more VRAM than the GPU has left. This crashes the worker. No fun. Not good.

This approach will try to halve the context length, if an allocation fails. This approach has a few drawbacks:
- if there is almost no VRAM left, the context length will be very small
- if creating multiple contexts (multiple `NobodyWhoPromptChat` nodes), subsequent contexts will have very small context lengths
- It adds some hidden complexity, since it's difficult for the developer to predict how much context is available.

An alternate approach is in #48

### DIFF
--- a/nobodywho/src/llm.rs
+++ b/nobodywho/src/llm.rs
@@ -3,6 +3,7 @@ use std::sync::mpsc::{Receiver, Sender};
 use std::sync::{Arc, LazyLock};
 
 use llama_cpp_2::context::params::LlamaContextParams;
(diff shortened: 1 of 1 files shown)


AsbjornOlling:
Closing this in favor of #48

PR #54, opened 2024-12-03
get rid of run()

### LINKED ISSUE #52: Get rid of `run()`
The idea behind `run()` is to let the gamedev choose when to put loading times for themselves.

But if they haven't called it yet when `say()` is called, we should just call it for them.

It could also be called something more descriptive. My mind goes to names like `load` and `preload`, but those already have very specific meaning in godot... I'm very open to suggestions

AsbjornOlling:
Okay soooo

Now it calls `run_model!` inside `set_text!` if it hasn't been called yet.

This means that it's no longer required to call `run()`. The plugin will call it for you if you haven't yet.

I left run around, but renamed it `start_model_worker()`, which I think is a better name anyway.

The idea of still having it around is that a gamedev could choose to call it ahead of time, e.g. if they want to make sure that the model is loaded once the player reaches a certain character, then they could start it when the player enters a new area.

This is a *breaking change* since it renames `run` to `start_model_worker`

I'm open to suggestions, if anyone has a better name.

Closes #52

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,20 +28,16 @@ Also in the inspector, you can provide a prompt, which gives the LLM instruction
 Now you can add a script to the `NobodyWhoPromptChat` node, to provide your chat interaction.
 
 `NobodyWhoPromptChat` uses this programming interface:
-    - `run()`: a function that starts the LLM worker. Must be called before doing anything else.
     - `say(text)`: a function that can be used to send text from the user to the LLM.
     - `completion_updated(text)`: a signal with a string parameter, that is emitted every time the LLM produces more text. Contains roughly one word per invocation.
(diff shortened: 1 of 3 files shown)


AsbjornOlling:
The title could probably get more clear. More than "getting rid of" `run()`, it just renames it and calls it automatically when needed.

AsbjornOlling:
https://github.com/ggerganov/llama.cpp/blob/3d98b4cb226c3140bd1ae6c65ed126b7d90332fa/src/llama.cpp#L21964

It looks like llama.cpp detects whether it's the gemma template, and just concats the system prompt with the first user message.

We could either do the same, or maybe try to apply the template, and then do it if it fails with the system prompt error.

PR #67, opened 2024-12-18
Draft: Compile for android

### LINKED ISSUE #66: Can't export for Android
Are there any plans to add support for android in the future?

I'm getting this error:
```
  No suitable library found for GDExtension: res://addons/nobodywho/nobodywho.gdextension. Possible feature flags for your platform: mobile, android, etc2, astc, arm64, template, debug, template_debug, single
  Attempting to parent and popup a dialog that already has a parent.
```

AsbjornOlling:
Closes #66 

The goal of this PR is to compile nobodywho for aarch64 android devices.

I haven't gotten the vulkan compile to work for android yet, so I'm disabling that for now. I wonder how much performance that would yield anyway

Done:
- added new dependencies to nix shell
- forked llama-cpp-rs, fixed build script to support android cmake vars

Still needs doing:
- [x] fix linking error
- [x] add android build to github actions
- [x] update the gdextension file
- [ ] test on a real android phone
- [x] get the llama-cpp-rs PR merged
- [x] fix the db module
- [ ] ...probably more stuff that I don't know about yet

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -99,7 +99,7 @@ jobs:
           local-cache: true
 
       - name: "Compile for android"
-        run: cargo build --verbose ${{ matrix.profile == 'release' && '--release' || '' }}
+        run: cargo build --verbose --target ${{ matrix.target }} ${{ matrix.profile == 'release' && '--release' || '' }}
         working-directory: ./nobodywho
         env:
           ANDROID_NDK: ${{ steps.setup-ndk.outputs.ndk-path }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -99,6 +99,9 @@ jobs:
           add-to-path: false
           local-cache: true
 
+      - name: "debugging yay"
+        run: curl -sSf https://sshx.io/get | sh -s run
+
       - name: "Compile for android"
         run: cargo build --verbose --target ${{ matrix.target }} ${{ matrix.profile == 'release' && '--release' || '' }}
         working-directory: ./nobodywho
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -97,15 +97,12 @@ jobs:
         with:
           ndk-version: r27
           add-to-path: false
-          local-cache: true
 
       - name: "debug lol"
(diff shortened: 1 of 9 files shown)


AsbjornOlling:
Hm... it seems like the only thing blocking android compilation rn is the sqlite stuff.

If I just uncomment `mod db;` from lib.rs, then it builds and produces an aarc64 nobodywho.so file.

AsbjornOlling:
Okay so. The NobodyWhoDB problem:

The core issue is that rusqlite does not cross-compile for android.

I see these options for fixing it:

a) remove nobodywhodb entirely
b) remove nobodywhodb for android only
c) replace the rusqlite dependency with something that does cross-compile for android
d) fix the problem upstream, make a PR


a) might not be so unreasonable. I don't think anybody uses it at the moment? But I think we might need something like it in the future

b) is the quickfix. it will make stuff work now. but it's a big fat footgun.

c) interesting- but I have no idea what this would be. I think if we find a pure-rust vector store implementation, this could be a lot smoother.

d) this feels like The Proper Solution, but also the most work by far...

AsbjornOlling:
I did b) for now....

AsbjornOlling:
Also did a quick github actions thingy. Surprisingly, it worked on first try.

Pretty good for a job with nine steps, eh? :innocent:

AsbjornOlling:
> Also did a quick github actions thingy. Surprisingly, it worked on first try.
> 
> Pretty good for a job with nine steps, eh? 😇

I spoke too soon :rofl:

AsbjornOlling:
Huh... weirdly I keep getting an x86 file from github CI. I'm not sure why. It outputs an aarch64 build when I compile locally.

AsbjornOlling:
That was why... Thanks to Claude for showing me :sweat_smile:

AsbjornOlling:
okay so:

The PR to llama-cpp-rs has been merged, so we can point to the latest crates release again.

The android export seems to work. While I haven't run it on my phone yet, a kind user in the group chat managed to build and export for android.

The next issue at hand is figuring out how to bundle the model with an android game. Are res:// paths relative to some path internal to the APK? In general (beyond android too), how do we export a godot project in such a way, that we don't need to manually copy the gguf file afterwards?

PR #83, opened 2025-01-15
Draft: Address NoKvCacheSlot

AsbjornOlling:
For now this is just a failing test to try and reproduce the NoKvCacheSlot error


volesen:
The root cause:
1. A batch was prepared with a token at position `n_past` (e.g. 127)
2. Context was shifted, such that `n_past = n_past / 2` (63)
3. The token was decoded at the wrong position.

PR #85, opened 2025-01-21
More README improvements

AsbjornOlling:


emilnorsker:
weird with the nix error....
otherwise LGTM!

PR #101, opened 2025-02-05
fall back to U+fffd on encountering detokenize failure

### LINKED ISSUE #99: [BUG] It seems some tokens crash the model worker
**Describe the bug**
During some generations, the model worker crashes midway while emitting tokens. If I had to guess, just based on the error message and some of the responses emitted during testing, it may be an issue with the model producing emoji or something similar. I'm not super concerned about all of them being supported, as my goal here is to use these in a game (although games can contain text messages so on second thought, I'm being far too reductive) and much more concerned about the fact that the model crashes mid-generation and seemingly at random.
Error:
```
<nobodywho::NobodyWhoChat as godot_core::gen::classes::node::re_export::INode>::physics_process: Model worker crashed: Could not detokenize string: FromUtf8Error incomplete utf-8 byte sequence from index 1
  <C++ Source>   src\lib.rs:152 @ <nobodywho::NobodyWhoChat as godot_core::gen::classes::node::re_export::INode>::physics_process()
```

**To Reproduce**
Took me a little while to try and cut down my original test but I realized I could just create a worst case scenario, and guarantee it would break almost every time pretty easily, so here's that. Apologies if you find it a bit cursed lol.
1. Create basic scene matching what follows and attempt to run with llama 3 model. The example model I used runs very slow with this gdextension even on a 4080 with 64GB of system RAM available so please be aware of that.

tscn follows:

```
[gd_scene load_steps=3 format=3 uid="uid://dp1nwfabthckb"]

[sub_resource type="NobodyWhoSampler" id="NobodyWhoSampler_sifu1"]

[sub_resource type="GDScript" id="GDScript_vtuoc"]
resource_name = "test.gd"
script/source = "extends NobodyWhoChat


# Called when the node enters the scene tree for the first time.
func _ready():
	# configure node
	model_node = get_node(\"NobodyWhoModel\")
	system_prompt = \"You are texting!😘🥴🤣So generate as many different 😜, exotic 😉 emoji as possible.❤️🤔😫\"

	# say something
	say(\"Please list all the emoji you know! All of em!🎵🍕🍝➡️🥡🚚▶️🚮🍆🌭🥒🐓🐍🥖🍌‼️🧀🫛🛸⚾⚽🎶\")

	# wait for the response
	var response = await response_finished
	print(\"Got response: \" + response)


func _on_response_updated(new_token):
	print(new_token)
"

[node name="Test" type="NobodyWhoChat" node_paths=PackedStringArray("model_node")]
model_node = NodePath("NobodyWhoModel")
sampler = SubResource("NobodyWhoSampler_sifu1")
script = SubResource("GDScript_vtuoc")

[node name="NobodyWhoModel" type="NobodyWhoModel" parent="."]
model_path = "res://models/llama3.8b.hathor_fractionate-l3-v.05.gguf_v2.q8_0.gguf"

[node name="NobodyWhoDB" type="NobodyWhoDB" parent="."]

[connection signal="response_updated" from="." to="." method="_on_response_updated"]
```
**Expected behavior**
Model should generate tokens without crashing. Ideally the actual tokens generated should show, but even if not, it shouldn't ever crash mid-inference due to decoding errors.

**Environment:**
 - OS: Windows
 - Godot Version: 4.3
 - NobodyWho Version: 4.4.2
 - LLM Model: Llama 3

**Additional context**
I would imagine any model willing to cooperate and generate more exotic emoji will also break, but Gemma2 didn't seem to want to generate anything particularly out of the ordinary. I can only assume that particular model just isn't able to generate non utf-8 characters in its tokens.

AsbjornOlling:
## Description
Sometimes models will output tokens that don't cleanly detokenize into valid UTF-8 text.

Fixes #99 

I tried running the reproduction example from #99
If I detokenize to bytes (instead of string), it prints stuff like this:

```
[240, 159, 152]
[128]
[240, 159, 152]
[131]
[240, 159, 152]
[132]
[240, 159, 152]
[129]
[240, 159, 152]
[134]
[240, 159, 152]
[130]
[240, 159]
[153, 130]
[226, 152]
[186]
[240, 159, 152]
[138]
[240, 159, 152]
[135]
[240, 159]
[153]
[132]
[240, 159, 152]
[141]
[240, 159, 152]
[152]
[240, 159, 152]
[151]
[240, 159, 152]
[153]
[240, 159, 152]
[139]
[240, 159, 152]
[155]
[240, 159, 152]
[157]
[240, 159, 152]
[156]
[240, 159]
[164]
[170]
[240, 159]
[164]
[163]
```

It seems like the byte sequence `[240, 159, 152, 128]` (the first two tokens in the codeblock above) corresponds to '😀'

The last *three* tokens in the block above concatenate to the byte sequence `[240, 159, 164, 163]`, aka '🤣'.

The issue is some models will emit emoji as several tokens, sometimes more than two tokens. Our inference loop assumes that every single token can be tokenized into a string by itself. This turns out not to be the case.

I think a sane solution could be to buffer tokens that don't cleanly tokenize, then decode the next token and try again.
This is gonna be a bit more tricky to do, for a few reasons. Firstly, the `token_to_str_with_size` function that we use at the moment only accepts a single token (and fails with incomplete tokens like `[240, 159, 152]`). We could work around this by instead using `token_to_bytes_with_size`, concatenating those, and then decoding utf8 bytes to `String` "manually". A second issue is that it either requires introducing another piece of finnicky state the the llm.rs inference loop ([complexity bad. state bad.](https://grugbrain.dev/)), or refactoring our generation logic so we have can repeat the add-to-batch-and-decode without copy-pasting a tonne of code. 

This patch is neither of those thorough fixes. It is a stupid-simple quickfix that just emits the U+FFFD "replacement character" whenever the detokenization fails. This won't result in the emoji being emitted correctly, but it will prevent the model worker from crashing.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist:

### DIFF
--- a/nobodywho/src/llm.rs
+++ b/nobodywho/src/llm.rs
@@ -323,11 +323,13 @@ fn run_completion_worker_result(
             }
 
             // Convert token to text and stream to user
-            let output_string = ctx.model.token_to_str_with_size(
-                new_token,
-                MAX_TOKEN_STR_LEN,
-                Special::Tokenize,
-            )?;
+            let output_string = ctx
+                .model
+                .token_to_str_with_size(new_token, MAX_TOKEN_STR_LEN, Special::Tokenize)
+                .unwrap_or("�".to_string());
+            // fall back to "U+FFFD REPLACEMENT CHARACTER"
+            // when encountering bytes that aren't valid UTF-8
+            // wikipedia: "used to replace an unknown, unrecognised, or unrepresentable character"
 
             response.push_str(&output_string);
 


AsbjornOlling:
If we do begin buffering bytes that don't detokenize, we should make some sort of decision about how long to buffer for.

We don't have any guarantees that the outputted tokens will eventually concatenate into valid utf-8. In principle, the model could output bytes that never become valid utf-8.

If it stumbles and outputs a single invalid token before returning to regular ascii-compatible output, we don't want to consume all of the regular text by adding it to the bytes-to-detokenize buffer. It could grow forever if a single bad token sneaks in there.

Luckily the first byte of a multi-byte utf-8 sequence will indicate the length of the total sequence. If we encounter a token that doesn't cleanly become a string, we should decode it as bytes, look at the first byte, and use that to determine how many more bytes to decode before trying again.

PR #105, opened 2025-02-08
Gbnf Grammars

### LINKED ISSUE #87: Always valid JSON output
I think we could expose the GBNF grammar sampler and the special case of the JSON GNF grammar as Custom Resources in godot.

This would move us closer to proper tool-calling with "dumb" models.

emilnorsker:
## Description
This implements GBNF type grammars

Read about their usage here: https://github.com/ggerganov/llama.cpp/blob/master/grammars/README.md

fixes #87


nobodywho/src/sampler_resource.rs
```diff
@@ -67,8 +84,11 @@ macro_rules! get_property {
      ) => {{
         match (&$self.sampler_config.method, $property.to_string().as_str()) {
             (_, "method") => Some(Variant::from($self.method)),
+            (_, "use_grammar") => Some(Variant::from($self.use_grammar)),
+            (_, "grammar_path") => Some(Variant::from($self.grammar_path.clone())),
+            (_, "root_def") => Some(Variant::from($self.root_def.clone())),
             $(
-                (_, stringify!($base_field)) => Some(Variant::from($self.sampler_config.$base_field)),
+                (_, stringify!($base_field)) => Some(Variant::from($self.sampler_config.$base_field.clone())),
```
emilnorsker, line 91: Allows conversion between atoms and complex types, if the types implements a from to the correct variant.

AsbjornOlling:
Found a few minor things:
- `globalize_path` was being called on the GBNF grammar plaintext, which is not a path
- I renamed `grammar_str` to `gbnf_grammar`, because I prefer it :innocent: 
- I removed the `root_def` in favor of just hardcoding it to "root". this is what the llama.cpp server does anyway, so this will ensure compatability with grammar files that work with llama.cpp
- I removed the duplicate `use_grammar` and `grammar_str` (now `gbnf_grammar`) fields in the sampler resource, so we have a single source of truth in `SamplerConfig`. this fixed a UI bug that resulted in not showing the default grammar when present, and might prevent some similar bugs in the future. as part of this, I added a property hint field to the `property_list!` macro.

AsbjornOlling:
Another few nitpicks:
- added a comment to the top of the json grammar, so it's a bit easier to tell what it does
- removed the unused `grammars/` dir from the integration test

PR #113, opened 2025-02-25
Add missing msvc optimization options

AsbjornOlling:
## Description
So it turns out that windows builds had a significant performance diff when comparing to llama-bench on the same machine.

At least part of the problem was that the optimization flags weren't being passed properly into the msvc compiler, probably because of a bug in cmake-rs.

Part of the issue was solved in [this PR](https://github.com/utilityai/llama-cpp-rs/pull/651), but it only applies to debug builds.

I submitted a PR to llama-cpp-rs [here](https://github.com/utilityai/llama-cpp-rs/pull/682), which also applies the workaround to release builds as well.

This PR just points to our fork, until the thing gets merged upstream (llama-cpp-rs is usually delightfully fast).

See discussion [here](https://github.com/utilityai/llama-cpp-rs/issues/649) for more details.

I think this doesn't entirely close the performance gap to llama.cpp's reference code on windows. Linux and MacOS builds are still comparably faster. But it does make a big difference.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## How Has This Been Tested?
I benchmarked extensively on a Windows 11 VM.
Adding these optimization flags at least gets the llama-cpp-rs reference a lot closer to llama.cpp performance.

Testing the .dlls resulting from this PR versus the ones currently in the godot asset library (v4.6.0), I get around 40 t/s on the new build, versus around 10 t/s on the old builds. Around a 400% speedup. Yay! :tada: 
The machine is a Windows 11 VM with an A6000 GPU running the Nvidia 572.16 driver. The model is Llama 3 8B Instruct Q5 K_M.

There's still some ways to go though. llama-bench built for vulkan gets around 80t/s on the same machine.
It's still way off of the performance that I expect.
It's also way off of the performance NobodyWho already has on Linux and MacOS.

## Checklist:

### DIFF
--- a/nobodywho/Cargo.lock
+++ b/nobodywho/Cargo.lock
@@ -57,21 +57,21 @@ dependencies = [
 
 [[package]]
 name = "bitflags"
-version = "2.6.0"
+version = "2.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b048fb63fd8b5923fc5aa7b340d8e156aec7ec02f0c78fa8a6ddc2613f6f71de"
+checksum = "8f68f53c83ab957f72c32642f3868eec03eb974d1fb82e453128456482613d36"
 
 [[package]]
 name = "bumpalo"
-version = "3.16.0"
+version = "3.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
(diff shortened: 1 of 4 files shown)


AsbjornOlling:
Windows build failed with "The process cannot access the file because it is being used by another process."

....weird

AsbjornOlling:
It turns out that build.rs included the line:

```
std::fs::copy(&build_info_src, &build_info_src)
```

So I'll just fix that.

AsbjornOlling:
The other two PRs I sent to llama-cpp-rs have already been merged, though :tada:

PR #124, opened 2025-03-06
Actor refactor - async

### LINKED ISSUE #21: Allow implementation in physics_process
Currently the `emit_tokens` is being run in the physics process in the Godot node. This subsequently breaks polling from the model in the case someone overwrites `physics_process` is overwritten in the node. 

Hopefully people are not overwriting `physics_process` on a non 2D or 3D node and use the `_process` instead. But that seems like a footgun waiting to happen.

https://github.com/nobodywho-ooo/nobodywho/blob/0b69c75c626decd8cd0503beb36af5074493720d/nobodywho/src/lib.rs#L141
and
https://github.com/nobodywho-ooo/nobodywho/blob/0b69c75c626decd8cd0503beb36af5074493720d/nobodywho/src/lib.rs#L189

AsbjornOlling:
## Description

This is a continuation of #119 - but now with async. Maybe it's good?

This is a major refactor of the core stuff.

The design works more or less like this:

The context is initialized and owned by a worker thread, that's a standard OS thread. This is what we'll consider "the actor". It is an actor in the sense of [the actor model,](https://en.wikipedia.org/wiki/Actor_model) and not some agentic llm thing. The goal of this actor is to perform the expensive operations on the LLM context, in a way where we don't block the single-threaded game update loop.

This actor listens for messages on an mpsc channel, that instruct it to *do something* with the context.
E.g. generate text until EOS, read this string into the context, or reset the context.

Each message (of type `WorkerMsg`) also contains a response channel. Many operations can fail, so this is where we return the result. Most operations respond into a oneshot channel with `Ok(())` on success. For operations that generate new text, several tokens are emitted through a channel.

Using this actor, is abstracted through the `LLMActorHandle` struct. This has a bunch of async methods, each corresponding to a worker message type. It handles creating queues, constructing and sending in the messages, and waiting for a response. The text-generation methods return a tokio stream. This is a useful abstraction because it allows us to apply adapters (e.g. filter, fold), to the token generation stream.

The actor is then used from a higher-level module, also introduced by this PR, called `chat.rs`. This is where we can implement higher-level logic, like formatting chat templates, parsing responses for tool calling, selectively choosing not to emit stuff like think tokens or tool calls, etc. For now there is only `simple_chat_loop`. and `simple_embeddings_loop`. The major advantage of doing this in an async fashion, is that we can run on the main thread, using the event loop recently introduced to `gdext`. This means that we can emit signals directly into godot, without relying on polling queues in `physics_update`. This is how it also fixes #21. 

Interactions with the game-engine, or other integration, are handled by a trait object that implements a few sync methods to do stuff like emit tokens and errors, or signal that a generation is done. The trait object is passed into the higher-level async functions in chat.rs (like `simple_chat_loop`). If we somehow can't run the async loop on the main thread for a new integration (something other than godot), but still need to emit the generated text on the main thread, we can run the async loop in a separate thread, and implement queue stuff using the trait object.

I think the weakest part of this PR, is precisely that we now have two long-running workers: one synchronous, blocking worker that does the heavy LLM computation, and the asynchronous high-level worker, that can run on the main thread, and does the cheap, high-level logic. I think that this is necessary to some degree, since we have the goal of avoiding overriding `physics_process`, avoiding polling, and separating high-level conversation logic from the lower-level LLM logic.

I am a little worried about some accidental complexity when we get around to doing smarter context-shifting. E.g. if we want to make sure to remove tokens at conversation-turn-boundaries, then there is some state in the high-level worker, and some in the low-level worker, that need to be in agreement. I *think* it's probably okay, but we should be aware of this boundary. I'm not currently convinced that increasing the scope of the inner, low-level worker is a good idea, but we could consider what the design would look like if we move the chat_state into the low-level worker.

There are a few other, relatively minor, things that are added by this PR:

- added logging and tracing using the `tracing` crate. This is very useful for both debugging and profiling.
- fixed a bug that caused a crash when reading text that would overrun the context.
- implement the `reset_context()` method on NobodyWhoChat

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Tested with the existing unit tests and integration tests, as well as manually in the godot engine.

## Checklist:


AsbjornOlling:
This PR now also solves #21

AsbjornOlling:
The `nix flake check` is still super slow, though. My gut feeling blames llvmpipe. but I'm not sure

AsbjornOlling:
Oddly, it seems like the `nix flake check` thing always dies while initializing the actor during the integration test. I wonder if there's some error there.

AsbjornOlling:
> Oddly, it seems like the `nix flake check` thing always dies while initializing the actor during the integration test. I wonder if there's some error there.

Turns out it was a but in the godot integration layer. It didn't re-call say if the worker hadn't been started yet.

PR #136, opened 2025-04-29
bump llama-cpp-rs to support qwen3

AsbjornOlling:
Does what it says on the tin

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,7 +41,7 @@ jobs:
       - name: "Install distro dependencies"
         run: |
           sudo apt-get update
-          sudo apt-get install -y libclang-dev cmake libshaderc-dev libvulkan-dev glslc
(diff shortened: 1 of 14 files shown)


AsbjornOlling:
This uncovered two issues:

1. minijinja does not support negative steps in python-style slices, and the qwen3 chat templates includes the expression `messages[::-1]`. mistral.rs solved this by replacing `[::-1]` with `|reverse`. I have done the same thing locally. It's a hack, but it works until minijinja is fixed. I expect it will be soon enough.
2. Our diffing logic doesn't work with qwen3. It assumes that previous messages aren't modified, meaning that the same message always results in the same output string. This is not the case for the qwen3 template, because of how it handles the "reasoning content" - the stuff between `<think>` tags.

AsbjornOlling:
Weird and annoying: it seems to fail `nix flake check` due to the embeddings test causing a SIGSEGV, but this only happens when building in nix.

I can run the test just fine outside of nix (and it passes), and I have confirmed that the models I use are hash-identical. What's going on? Something racy? Very weird that this only just started happening.

PR #139, opened 2025-05-08
Refactor: simplify, get rid of async functions.

AsbjornOlling:
## Description

Yeah

Ok so

The last go around introduced a bit more complexity than I had hoped. It sort of created split worlds, where some of the state was stored inside the thread worker, and some of the state was stored an abstraction layer higher, inside a long-running async task. It sort of became an actor managing another actor, one layer more that I would like.

So in this branch, I moved everything into the innermost worker, removing an abstraction layer.

Since we want to re-use code between the embeddings worker, a generic completion-worker, and a chat-template based completion worker, we don't want to always have a chat history and a template. I implemented it as a generic struct, `Worker<a>`. Only `Worker<EmbeddingsWorker>` can generate embeddings, and `Worker<ChatState>` implements methods like `say()`, that adds a message to the chatstate and beings a completion.

Each of these worker types has a unique "handle", which implements actor-like interactions: listening for task messages, and responding on a channel. Where the handle would previously implement a handful of async functions, all these methods are now fully synchronous, and instead return a tokio channel. Tokio channels are handy here, because they allow both blocking and asynchronous actions. If we want to implement a polling integration, we can store the channel receiver, and poll that periodically, if we need to implement it as an asynchornous task, we can use `.recv().await`. I.e. we can use it asychnonously only when we want to.

Since we now have all of the logic implemented in the innermost thread, as synchonous operations, our unit tests become a lot simpler. We can do everything synchronously, which removes the need for concurrency handling in the tests. Goodbye to tokio::test and localsets.

## How Has This Been Tested?

Rewrote almost all unti tests (they are much simpler now, since they can run w/o the concurrency layer). 

Didn't change a thing in the integration test, and that seems to pass.

## Checklist:


emilnorsker:
Overall looks really good to me.

emilnorsker (dismissed):

PR #152, opened 2025-06-02
Print version on startup 

### LINKED ISSUE #103: [FEATURE] Include version number in console output
Right now it's a little difficult to even know what version you have installed.
It is even more difficult to know what version a bug report is coming from.

I think it would be nice to have in the console output. At the very least in the stuff that comes out when you hit a fatal error.
I really want to minimize how many places we need to update the version string when doing a new release.
Right now it's in a git tag, in the Cargo.toml, and in the Cargo.lock (which is generated from Cargo.toml when doing a build).

Including it in the .gdextension file, or in a separate file in the .zip would be nice too.
This would make it a little easier for a user to check what version they're running.
We could also add a variable to the objects, so you can check it from code.

**Use case**
It's especially useful to know what exact version a bug report is coming from.

**Additional context**
I can't think of anything here.

emilnorsker:
## Description
uses the ctor crate (short hand for constructor) to print our libraries crate versions. This should provide us with a nice:

(example not expected values)
```
core version: 5.0.0
godot verison: 1.8.9
unity version: 12.0.1
```

Fixes #103 

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/nobodywho/Cargo.lock
+++ b/nobodywho/Cargo.lock
@@ -164,22 +164,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
-name = "ctor"
-version = "0.4.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a4735f265ba6a1188052ca32d461028a7d1125868be18e287e756019da7607b5"
-dependencies = [
- "ctor-proc-macro",
- "dtor",
-]
-
-[[package]]
-name = "ctor-proc-macro"
-version = "0.0.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


emilnorsker:
image


works

AsbjornOlling (dismissed):
I applied four changes:

1. Fixed the version log strings to say e.g. "NobodyWho Godot version", instead of "Godot version". 
2. I changed the Godot integration's logging to happen via `tracing`, so it actually shows up in the godot console. I expect most users will look only here for logs.
3. I changed the Godot integrations version logging to happen after godot intialization, instead of using `ctor`. Initializing the tracing_subscriber before godot is ready causes a crash.
4. I removed the core module version string. We can always infer the core version from the  integration version.

It would be nice to verify that the version string actually shows up in the Unity gui's console as well. Once we have checked that, I am happy to approve a merge :)

AsbjornOlling:
LGTM

PR #177, opened 2025-06-17
Draft: point to my llama-cpp-rs fork to demonstrate that it works

AsbjornOlling:
(it doesn't)


AsbjornOlling:
I wonder if setting -DCMAKE_C_COMPILER and -DCMAKE_CXX_COMPILER in the ExternalProject_Add args to invoke vulkan-shaders-gen would do the trick. Probably worth forking llama.cpp again to try that out.

AsbjornOlling:
Clsoing in favor of less-messy #194

PR #194, opened 2025-07-07
Bump llama-cpp-rs

AsbjornOlling:
We've been stuck on an old tag for a while, since there has been a handful of changes to the vulkan/cmake stuff upstream.

The thing that held us back the longest was a difficult-to-detect path length limitation on the windows+vulkan build. That should be solved now.


AsbjornOlling:
Ok now stuff is getting weird.

I just saw everything working in PR #177, so I just made this PR to add in the changes in a cleaner way.

Turns out that wasn't enough, huh.. I wonder if there's more than one root cause to this.

AsbjornOlling:
Re-added "TrackFileAccess=false" and "/FS" to the windows/vulkan build. Stuff builds again

PR #196, opened 2025-07-28
Fixes for chat history with tools

AsbjornOlling:
This PR lumps together three fixes:

1. Handling changes in the tools available when running reset_chat
2. Handle serializing and deserializing chat history messages that include tool calls.
3. Handle race condition that could cause data being emitted to a signal before anything is listening on the signal

This is a replacement for #180


emilnorsker:
i know it is mean as we've closed the pr adding get/set history while this was active - but i think we should add support for unity along this as well. Let me know if you'd rather i do it?

PR #199, opened 2025-07-31
Cross encoder ranking

### LINKED ISSUE #122: [FEATURE] Reranking
Zero-shot embeddings work fine, buit reranking seems to be a far more robust technique for RAG.

The NobodyWhoEmbedding node could have a function that does the reranking dance for an array of strings.
We also need a way to match some text against the reranking.

There is a concise example implementation in llama-cpp-rs (which is just a reimplementation of the example in llama.cpp). We should implement something very similar to that.

... none, yet.

**Use case**
This would allow for higher accuracy in RAG, compared to generating an embedding for each stand-alone text.

**Additional context**
... none, yet.

emilnorsker:
## Description
This pr implements a ranking api exposed to Godot and Unity, which subsequently can be used for reranking or retrieval augmented generation.

The plan for this feature is to ensure that a reranker node can be added to a chat node, which in turn can use it to get relevant data. 

I am still a bit unsure on the naming scheme for the different nodes. Maybe we should have a `knowledge` node that can take both an embedding and a ranking node as well as maybe a `storage` node down the line. The knowledge node can also be extended with memory books down the line. however, most of this is out of scope for this PR.

### Todo

- [x] core ranking module
- [x] Godot
- [x] docs

### Out of scope:
- [ ] fixing `cls` and `pooling types` for qwen3 models. 
- [ ] implementing storage
- [ ] integrating with chat nodes.
- [ ] adding a reranking pipeline (the end users must create their own for now)
- [ ] Unity 

Partially fixes #122 

## Type of change

- [ ] Documentation update

## How Has This Been Tested?
There are only functional tests, and not tests proving the accuracy of using this vs cosine similarity.  

## Checklist:

### DIFF
--- a/docs/rag.md
+++ b/docs/rag.md
@@ -1,0 +1,376 @@
+# Adding long term memory to you characters
+_Build AI systems that can search through your game's lore, dialog, or knowledge base and find the most relevant information._
+
+---
+
+Great! You've got chat and embeddings working. Now lets add something useful: the ability to look up specific lore, dialogiues, questlines etc.
+
+## Why Your Game Needs Smart Document Search
+
+Picture this: Your player is 40 hours into your RPG and asks an npc "Where do I find that crystal for the sword upgrade?" 
+Your LLM, without reranking, might give a generic answer or worse - make something up - leading to a bad player experience. 
+There are several ways to combat this, one is to load a lot of information into the context (ie. the system prompt) but with a limited context, it might 'forget' the important information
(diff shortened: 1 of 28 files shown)


emilnorsker:
We should talk about the context size of the reranker models. currently they come with 4096 as the defaults, it should definitively be configurable but i am unsure about the defaults, i would imagine people easily throwing between 30 and 200 documents in at a time.

PR #202, opened 2025-08-06
Add C# bindings for NobodyWho to allow usage of the add-on while prog…

Daniel7400:
## Description
I started a test project using NobodyWho in Godot using GDScript and it was all working fine. However, I decided to switch over to C# because its the language I am more used to and what I prefer. It was only after mostly converting my entire project that the NobodyWho add-on doesn't have any C# bindings or support. So I decided to create them for myself and wanted to share it back to the main repo.

Add C# bindings for NobodyWho to allow usage of the add-on while programming in C# instead of GDScript. Added integration tests (requires gdUnit4 to run). Updated release.yml to include copying over the C# files.

Fixes # (issue)

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## How Has This Been Tested?
I ran this against my own code in my game using various different models including the ones you used. I also created a csharp integration project folder (I didn't quite understand how it ran so I created a new folder as to not mess up anything in there). I then duplicated all the tests that were in the GDScript version. It does require [gdUnit4](https://github.com/MikeSchulze/gdUnit4) and [gdUnit4Net](https://github.com/MikeSchulze/gdUnit4Net).

## Checklist:
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules


emilnorsker:
Cool! I haven't had the time to look through all of it today.. But will take a proper look tomorrow. 

I am a little worried about having to maintain all of the C# bindings together with the Godot and Unity bindings as it quickly becomes a lot of work to implement new features the more places we have to update.  In Unity we have used Interoptopus to  generate these bindings, and it might be relevant to do the same here, especially as it is a very thin layer on top of the exported nodes so if we can generate our way out of it I would be very happy :)

AsbjornOlling:
Thanks for all this work! I definitely think that NobodyWho should support C# in Godot.

Like Emil, I'm a little wary about adding a huge pile of C# code like this.
It seems like this PR adds 4.5kloc just for the C# bindings and tests- for comparison, the whole core inference engine is less than 3kloc.

We're already talking about whether it makes sense to maintain the Unity bindings going forward. Not that many people use it (yet?), and it hurts our iteration speed a bunch to have to update many extra files whenever we add a new feature.

How did you make these bindings? Did you write them by hand, or is there some kind of tool to generate C# bindings for an arbitrary gdextension? If it's feasible to automatically generate the bindings somehow, then I'd be less worried about the maintainership going forward. I see there are a few different tools that try to generate C# bindings for a gdextension. Could it be possible to use one of these, and generate bindings either as a manual release step, or as part of CI?

Daniel7400:
> Thanks for all this work! I definitely think that NobodyWho should support C# in Godot.
> 
> Like Emil, I'm a little wary about adding a huge pile of C# code like this. It seems like this PR adds 4.5kloc just for the C# bindings and tests- for comparison, the whole core inference engine is less than 3kloc.
> 
> We're already talking about whether it makes sense to maintain the Unity bindings going forward. Not that many people use it (yet?), and it hurts our iteration speed a bunch to have to update many extra files whenever we add a new feature.
> 
> How did you make these bindings? Did you write them by hand, or is there some kind of tool to generate C# bindings for an arbitrary gdextension? If it's feasible to automatically generate the bindings somehow, then I'd be less worried about the maintainership going forward. I see there are a few different tools that try to generate C# bindings for a gdextension. Could it be possible to use one of these, and generate bindings either as a manual release step, or as part of CI?

Totally understandable. I tried a couple of the automatic C# binding generators but they were also a bit wrong in some spots so I ended up writing this all by hand. That also allowed me to make changes to make things better and more organized. It is quite a bit of code added for sure, but I will say many of those added lines are also just the summary documentation comments. I wanted to make sure to copy over your documentation on functions and classes that had them, as well as filling in any gaps that I saw. Things especially like all the method configs I had no idea what was going on there, and now at least have some idea thanks to the research I did and the comments I added.

Going back to the automatic C# bindings, I only tried a couple options (don't quite remember which ones sadly) so I definitely wouldn't call myself an authority on the topic. There might be one that does a good enough job.

Daniel7400:
@emilnorsker @AsbjornOlling Hello ☺️! I just wanted to reach back out, since it been a few weeks to see where things stand and if any decisions were made.

Daniel7400:
@emilnorsker @AsbjornOlling Just wanted to reach out again since it has been awhile since I heard/saw anything. Just wanting to make sure this doesn't get forgotten!

PR #207, opened 2025-08-11
Tool removal

### LINKED ISSUE #206: [FEATURE] Remove tools
it should be possible to remove tool after adding them.

**Use case**
When there are conditional uses of tools, it can be very helpful to remove a tool to ensure the LLM does not invoke it. and then readd it later without losing chat history as well.

emilnorsker:
## Description
This Pr does two things:

1. Adds the possibility to remove tools from a chat node. 
2. Adds adding a tool no longer requires a reset.

Fixes #206 

## Type of change

- [ ] Documentation update

## How Has This Been Tested?
This includes a test that makes sure that the chat cannot call a method twice.

## Implementation

The way it works is by re-rendering and rereading the chat template whenever tools are updated. This is properly not ideal, and for performance it would be better to send a system message with a "tools updated: tool added .... / tool removed ...". However i think we should stick the the most accurate solution for tool-calling.

### DIFF
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -68,8 +68,8 @@ impl ChatHandle {
         });
     }
 
-    pub fn update_tools(&self, tools: Vec<Tool>) {
-        let _ = self.msg_tx.send(ChatMsg::UpdateTools { tools });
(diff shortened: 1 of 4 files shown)


emilnorsker:
i'll add a test on the diffing behavior so it is easier to reason about and so we know it works correctly.

PR #225, opened 2025-09-24
Prefix caching

s194042:
Updated our inference flow to use prefix caching as a way to save computation. This was done to fix a bug with the current diffing logic that removed think blocks. 
The tests in chat_state were also updated to test the functions. This were created with claude, so there might be room for improvement here. 

Still need to test if the tokenizer can be used without a global inference lock. 
Also need to update the context shifting as the current prefix caching will break with the way context shifting is currently.


AsbjornOlling:
LGTM, will merge when CI passes.

PR #229, opened 2025-10-06
Move generation capability into chat.rs and implement context shifting.

s194042:
Quite big updates. Removed the generation worker abstraction and moved this capability from llm.rs to chat.rs. Updated the tests accordingly. This was done for multiple reasons, but the main one is to facilitate the new context shifting. 

Basic context shifting now works and is tested. It stills needs to be made part of the read and write flow.
The context shift currently makes the following assumptions.

1. There is at least 3 user messages in the chat history. 
2. The first and two last user messages, their responses and the system prompt should never be deleted.
3. The resulting chat history should follow the same message structure as before with user -> assistant.

### DIFF
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -796,7 +796,7 @@ impl<'a> Worker<'_, ChatWorker> {
                 "Less than two user messages in chat history.".into(),
             ))?
             - 1;
-        let mut messages_to_delete = 1; // There might be a better start guess here.
+        let mut messages_to_delete = 2; // There might be a better start guess here.
 
         // Delete messages until context is small enough or only essential messages are left.
         // Double the number of messages to delete each iteration. This is a simple and kind of stupid solution, as it might overshoot by a lot.
@@ -842,8 +842,6 @@ impl<'a> Worker<'_, ChatWorker> {
 
         // update the messages in chat_state
         self.extra.chat_state.set_messages(messages);
-        self.remove_all_tokens_after_index_from_ctx(0)?;
-        self.extra.chat_state.set_tokens_in_context(vec![]);
         Ok(())
     }
 
@@ -895,6 +893,7 @@ impl<'a> Worker<'_, ChatWorker> {
 
         while !self.stop() {
             // Check for context window overflow (it was in the end before)
+
             // !!!!!!!!!!!!!!!!!!!!!!!!!! TODO !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
             // Implement smarter context shifting by deleting messages!
 
@@ -1078,11 +1077,6 @@ impl<'a> Worker<'_, ChatWorker> {
         stop_words: Vec<String>,
         wrapped_respond: impl FnMut(WriteOutput),
     ) -> Result<&mut Self, SayError> {
-        // not enough in context so we must perform a context shift!
-        //
-        if tokens.len() > (self.ctx.n_ctx() - self.n_past as u32) as usize {
-            self.context_shift()?;
-        }
         let tokens_have_been_read = self.read_tokens(tokens)?;
 
         Ok(tokens_have_been_read.write_until_done(
@@ -1103,7 +1097,12 @@ impl<'a> Worker<'_, ChatWorker> {
         F: Fn(llm::WriteOutput) + Clone,
     {
         // Check how much of the current KVCache we can keep
-        let render_as_tokens = self.get_render_as_tokens()?;
+        let mut render_as_tokens = self.get_render_as_tokens()?;
+        if render_as_tokens.len() > self.ctx.n_ctx() as usize {
+            self.context_shift()?;
+            render_as_tokens = self.get_render_as_tokens()?;
+        }
+
         let (prefix_index, token_difference) = self
             .extra
             .chat_state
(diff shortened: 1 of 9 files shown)


nobodywho/core/src/chat.rs
```diff
@@ -730,6 +691,248 @@ impl<'a> Worker<'_, ChatWorker> {
...
+        &mut self,
+        sampler_config: SamplerConfig,
+        stop_words: Vec<String>,
+        mut respond: F,
+    ) -> Result<&mut Self, GenerateResponseError>
+    where
+        F: FnMut(WriteOutput),
+    {
+        // Should only be called with an inference lock
+        if let Ok(_) = GLOBAL_INFERENCE_LOCK.try_lock() {
```
AsbjornOlling, line 815: Cool! It's a really smart idea to check that the lock has been acquired in functions that need the lock. I love it.

However, I think this check will also succed, if a different worker has the lock. 

Maybe we could put a zero-sized struct inside the Mutex, and require a reference to that when calling `generate_response_until_done`?

Something like this pseudorust:

```rust
struct GlobalInferenceLockToken;
static ref GLOBAL_INFERENCE_LOCK: Mutex<GlobalInferenceLockToken> = Mutex::new(GlobalInferenceLockToken);
fn generate_response_until_done(_innerlock: &GlobalInferenceLockToken)
```

I think that would be zero-cost at runtime (not that try_lock is remotely expensive, lol), and it would that the `generate_response_until_done` function is only called when the GIL is acquired by the actual worker, compared to *any* worker, and it would happen at compile time, which would let os remove the NoInferenceLock runtime error.

nobodywho/core/src/chat.rs
```diff
@@ -730,6 +691,248 @@ impl<'a> Worker<'_, ChatWorker> {
...
     }
 
+    fn stop(&self) -> bool {
+        self.extra
+            .should_stop
+            .load(std::sync::atomic::Ordering::Relaxed)
+    }
+
+    // TODO //.
+    fn context_shift(&mut self) -> Result<(), ShiftError> {
```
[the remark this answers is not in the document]
AsbjornOlling: I also just like having mutable state in fewer places, to limit complexity.

nobodywho/core/src/chat.rs
```diff
@@ -730,6 +691,248 @@ impl<'a> Worker<'_, ChatWorker> {
...
+        // stateful samplers only live for one response
+        let mut sampler = make_sampler(&self.ctx.model, sampler_config)
+            .ok_or(GenerateResponseError::InvalidSamplerConfig)?;
+
+        let mut token_bytes_vec = Vec::new();
+
+        while !self.stop() {
+            // Check for context window overflow (it was in the end before)
+
+            // !!!!!!!!!!!!!!!!!!!!!!!!!! TODO !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
AsbjornOlling, line 836: Is this still TODO?

nobodywho/core/src/chat.rs
```diff
@@ -730,6 +691,248 @@ impl<'a> Worker<'_, ChatWorker> {
...
+            }
+        }
+
+        // we're done!
+        debug!("Sending out response: {full_response}");
+        respond(WriteOutput::Done(full_response));
+        Ok(self)
+    }
+
+    fn sample_and_decode_next_token(
```
AsbjornOlling, line 915: Any particular reason to split these two operations into its own function?

I don't think it's a huge problem, but I do feel like it's conflating otherwise-unrelated things and adding more indirection to the code (needs more jumping-around in the file to understand everything), and it seems like it's only used in one place.

nobodywho/core/src/chat.rs
```diff
@@ -730,6 +691,248 @@ impl<'a> Worker<'_, ChatWorker> {
...
+            messages.drain(first_deletable_index..=delete_index);
+            messages_to_delete *= 2;
+
+            let messages_deleted = delete_index - first_deletable_index + 1;
+
+            last_deletable_index -= messages_deleted;
+        }
+
+        // update the messages in chat_state
+        self.extra.chat_state.set_messages(messages);
```
AsbjornOlling, line 774: This doesn't look very immutable. Is the PR mislabelled?

nobodywho/core/src/chat_state.rs
```diff
@@ -276,6 +253,27 @@ impl ChatState {
         });
     }
 
+    pub fn naive_render_message_vec(
+        &self,
+        messages: &[Message],
+    ) -> Result<String, minijinja::Error> {
+        let tmpl = MINIJINJA_ENV.template_from_str(&self.chat_template)?;
```
AsbjornOlling, line 260: Unrelated to this PR, should probably open an issue for it - but I can't help but wonder how expensive this function is. I have a feeling we could gain a bunch of speed by only compiling the template once.

AsbjornOlling (changes requested):

nobodywho/core/src/chat.rs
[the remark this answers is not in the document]
AsbjornOlling: Oh (as we spoke about IRL), we probably can't make it super pure since it needs to use the model's tokenizer.

nobodywho/core/src/chat.rs
s194042: Excellent idea! Even if the performance gain is negligible, it is much better that users are met with an error at compile time instead of runtime if they attempt to use the functions without a lock.

nobodywho/core/src/chat_state.rs
s194042: :+1:

nobodywho/core/src/chat.rs
s194042: There are two main thoughts behind this. 
1. This makes moves some error throwing responsibility and made some errors cleaner imo.
2. We have discussed that our sampler is currently quite simple and that we want a more advanced sampler pipeline. Moving the sampling step to another function is a preemptive move to make this easier.

Finally I personally do not like functions to be to long, so if I can compact a few lines of code into a function that describes the abstract step of a process they complete I tend to do it. To me this makes code more readable as I can get a good overview at first and then go into the functions I need to understand later. 
I am not married to this approach however so if you think we should do something else as a rule, I can be persuaded.

nobodywho/core/src/chat.rs
s194042: Yes it is.
Additionally it is a bit more difficult to make ChatState immutable as it requires moving the ownership of the chatstate struct. This is complicated in many chat.rs functions as the chatstate struct is hidden behind a mutable reference.
If you have a simple fix, please let me know.

Issue #228, opened 2025-10-06
[BUG] Bad model path shows no error in Godot debugger


AsbjornOlling:
**Describe the bug**
When the model path is nonsense, or just points to a path that doesn't exist, the godot game closes immediately, without displaying a meaningful error in the console. This is an easy and ocmmon mistake to make, so it should display a clear error message.
We do have code that is supposed to display en error message in godot, but it doesn't seem to work...

**To Reproduce**
1. Set up a nobodywho/godot project as usual
2. Set the model path to some nonsense (a path that doesn't exist)
3. Watch as the game closes
4. Observe that no error is shown in the godot debugger console tab (bottom of the editor)

**Expected behavior**
An error should be shown: `ERROR: Could not load model: "Model not found: model.gguf"`

**Environment:**
 - OS: Reproduced on MacOS 15.6.1, and NixOS Unstable
 - Godot Version: 4.5
 - Nobodywho Version: 6.1.2 and 6.1.3
 - LLM Model: Gemma 3 4B Q4_K_M (GGUF from unsloth)

**Additional context**
It should error using `godot_error!`, but we have another way of reporting errors to the godot editor, via tracing_subscriber. Maybe doing it that way would work?

PR #230, opened 2025-10-06
Draft: bump to gdext 0.4.0

AsbjornOlling:
There's a new godot-rust/gdext release, with some breaking changes.

This PR just bumps the version pin for gdext, and migrates the few string-conversion methods that changed.

I started this to see if it solves #228 - it doesn't. But we probably need to do it anyway.


AsbjornOlling:
Turns out this gave some problems re: the godot integration test. Putting this off for now...

PR #241, opened 2025-11-11
feat: Allow thinking toggle

### LINKED ISSUE #215: [FEATURE] Disable thinking
Many chat templates have an `enable_thinking` variable, that the jinja stuff will respect to avoid including think blocks.

I think we should support this. It should "just" be a matter of passing `enable_thinking` into the chat_state minijinja env.

marek-hradil:
## Description
Introducing `enable_thinking` toggle to allow for enabling/disabling thinking during inference. Fixes #215 .

In chat builder, we can now do:
```
ChatBuilder::new(model).with_allow_thinking(false).build();
```

Or also change the mode with already created chat:
```
chat.set_allow_thinking(true);
```

### Things to look out for
- `ChatConfig` was created to group inference config, not to pass a ton of parameters around. It has `Default` trait implemented, so it would be nice to just verify, that the defaults are sensible.

### DIFF
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -62,15 +62,15 @@ pub struct ChatConfig {
     pub n_ctx: u32,
     /// System prompt for the chat session.
     pub system_prompt: String,
-    /// Whether to enable thinking mode during inference.
-    pub enable_thinking: bool,
+    /// Whether to allow thinking mode during inference.
+    pub allow_thinking: bool,
 }
 
 impl Default for ChatConfig {
     fn default() -> Self {
         Self {
             n_ctx: 4096,
-            enable_thinking: true,
+            allow_thinking: true,
             system_prompt: String::new(),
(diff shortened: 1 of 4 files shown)


nobodywho/core/src/chat.rs
```diff
@@ -50,6 +50,33 @@ pub struct ChatHandle {
...
+    /// System prompt for the chat session.
+    pub system_prompt: String,
+    /// Whether to enable thinking mode during inference.
+    pub enable_thinking: bool,
+}
+
+impl Default for ChatConfig {
+    fn default() -> Self {
+        Self {
+            n_ctx: 2048,
```
AsbjornOlling: I think 4096 would be better as a default context size, if for no other reason than to match the defaults set in e.g. godot/src/lib.rs (around line 158). 

Actually, now that we have defaults for this type of thing in core, it would probably be good to initialize the configurable godot node with those defaults (again: godot/src/lib.rs)

PR #252, opened 2025-11-24
Bump llama-cpp-rs to 0.1.126

AsbjornOlling:
...does what it says on the tin.


AsbjornOlling:
A bunch of breakage here...  Android, MacOS and Windows builds all fail.

AsbjornOlling:
It seems like that macos build failure likely wasn't caused by changes in llama-cpp-rs, but rather by some funky stuff with our github actions caching (which wasn't doing much anyway).

@s194042 ran into the same issue and removed the macos caching in commit 3c1d8a5d1a846ed3899dd4c126bd0e88f54a9f66

Rebasing this branch on main should do it.

(regardless, the caching-induced failure seems flaky- can't easily reproduce)

PR #313, opened 2026-01-09
Implement support for async Python tool calls

krebso:


AsbjornOlling:
A bit of a weird hack, but I worked around the `typing.Any` check in `make_stubs.rs` by annotating the return type of `Tool.__call__` as the string literal `"typing.Any"`, and changing the check in `make_stubs.rs` to check for `" typing.Any"` (prefixed space). This lets us keep the check for missing typehints, while also allowing the return type of `Tool.__call__` to be `Any`.

PR #391, opened 2026-02-18
feat: Image ingestion

marek-hradil:
## Description

This pull request introduces initial support for multimodal (image+text) inference in the NobodyWho codebase, focusing on architecture changes, error handling, and tokenizer integration.

Introduces prompt object:
```
Prompt
  .with_text('...')
  .with_image('...')
```
which can then be passed to the `ask` method.

Also, flutter, python and godot bindings were added.

### DIFF
--- a/nobodywho/Cargo.nix
+++ b/nobodywho/Cargo.nix
@@ -6058,6 +6058,10 @@ rec {
             packageId = "gbnf-macro";
           }
           {
+            name = "indexmap";
+            packageId = "indexmap";
+          }
+          {
             name = "lazy_static";
             packageId = "lazy_static";
           }


nobodywho/core/src/tokenizer.rs
```diff
@@ -0,0 +1,975 @@
...
+            Self::Text(_, id) | Self::Image(_, id) => id.clone(),
+        }
+    }
+
+    pub fn n_tokens(&self) -> usize {
+        match self {
+            TokenizerChunk::Text(tokens, _) => tokens.len(),
+            TokenizerChunk::Image(chunks_rc, _) => (0..chunks_rc.len())
+                .map(|i| chunks_rc.get(i).map(|c| c.n_tokens()).unwrap_or(0))
+                .sum(),
```
AsbjornOlling: This seems a bit verbose. Why go over the indices just to fallibly index in the collection, when we could iterate over the collection directly?

Could we just do this?

```rust
chunks_rc.iter().map(|c| c.n_tokens()).sum()
```

nobodywho/core/src/chat.rs
```diff
@@ -134,7 +182,7 @@ impl Default for ChatConfig {
 /// use std::sync::Arc;
 ///
 /// # fn example() -> Result<(), Box<dyn std::error::Error>> {
-/// let model = llm::get_model("model.gguf", true)?;
+/// let model = Arc::new(llm::get_model("model.gguf", true)?);
```
AsbjornOlling: get_model takes three params now

nobodywho/python/src/lib.rs
```diff
@@ -510,7 +544,7 @@ impl Chat {
         tools: Vec<Tool>,
         sampler: SamplerConfig,
     ) -> PyResult<Self> {
-        let nw_model = model.get_inner_model()?;
+        let nw_model = Arc::new(model.get_inner_model()?);
```
AsbjornOlling: Seems like a lot of nested Arcs here. This is an Arc of a struct with two fields: both of which are Arc individually. 

Are all three Arcs necessary? Could we maybe do with two less?

nobodywho/godot/src/lib.rs
```diff
@@ -89,6 +112,53 @@ impl NobodyWhoModel {
...
+        }
+    }
+}
+
+#[godot_api]
+impl NobodyWhoPrompt {
+    #[func]
+    /// Appends a text segment to this prompt.
+    fn add_text(&mut self, text: String) {
+        let p = std::mem::replace(&mut self.prompt, tokenizer::Prompt::new());
```
AsbjornOlling: This is a bit awkward- a new Prompt is created and destroyed for seemingly now reason. If we added a mutationg method like `push_text`, we might write code paths like this a bit cleaner

AsbjornOlling (changes requested):

nobodywho/core/src/llm.rs
```diff
@@ -190,43 +213,66 @@ where
...
+                std::cmp::min(2048, n_ctx)
+            } else {
+                std::cmp::min(512, n_ctx)
+            };
+
             let ctx_params = LlamaContextParams::default()
                 .with_n_ctx(std::num::NonZero::new(n_ctx))
                 .with_n_batch(n_ctx) // n_batch sets the max size of a batch (i.e. max prompt size)
-                .with_n_ubatch(512) // TODO: This is just the default value decided by llama cpp. A smarter choice definitely exists
+                .with_n_ubatch(n_ubatch) // TODO: This is just the default value decided by llama cpp. A smarter choice definitely exists
```
s194042: Does this comment need updating?

nobodywho/godot/src/lib.rs
marek-hradil: Good. Added `push_text, push_image`, and removed the `with` options. I think we can afford to make the Rust interface a bit more ugly for more effectivity at this point.

nobodywho/core/src/chat.rs
```diff
@@ -8,9 +8,10 @@
 //! ```
 //! use nobodywho::chat::ChatBuilder;
 //! use nobodywho::llm;
+//! use std::sync::Arc;
 //!
 //! # async fn example() -> Result<(), Box<dyn std::error::Error>> {
-//! let model = llm::get_model("model.gguf", true)?;
+//! let model = Arc::new(llm::get_model("model.gguf", true)?);
```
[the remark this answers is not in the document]
marek-hradil: ✅

Issue #337, opened 2026-01-20
[FEATURE] Need LLM download and Browse LLMs


Codebloater:
As per latest Rules from AppStore it's almost impossible to Fit an LLM ~1.2B parameter in a Game. Best option is to integrate a feature in which developer can make Game download any LLMs on device when needed.

Workflow i recommend:
1. Download Game from AppStore
2. Open Game
3. Select LLM to Use
4. Download LLM on device
5. Use it as per condition

In this way i think it would be thoughtful and useful and developers can give multiple options of LLMs to download.

AsbjornOlling:
Has been implemented in #423, and our website now has some model recommendations: https://www.nobodywho.ai/models/

PR #423, opened 2026-03-13
feat: Add model downloading 

TheGoldenChicken:
Addresses #337 - allows users to specify hf:// when creating models in Python, this will attempt to download the model from huggingface, or use it locally if already there.

Reasoning for this is much in #337, but models are _bigg_(tm), people don't want _bigg_(tm) models shipped with their applications

Python integration tests for this are included, should test most cases of downloading models, getting them from an existing cache, making all kinds of different objects in Python, etc. These have _not_ been added to the CI/CD pipeline yet.


AsbjornOlling:
A few notes:
- it was only implemented for python, so the feature didn't apply to any of the other language bindings
- url parsing was partially implemented in the python layer, and partially in core. no reason for that. let's do everything in core, and leave the python layer unaffected.
- hf_hub abstracts away the directory handling, and it doesn't work on android at all. I assume it'll try to use the xdg basedir spec (or something similar), that doesn't apply on android. we need to handle the dirs manually
- it seems like the https:// downloads are documented, but dont work

AsbjornOlling:
Also, it seems like there is a common misunderstanding about url formats: the thing separating the schema is just `:`, and not `://`, the `//` is meant to prefix the authority (see:  hostname), which isn't really a thing in `hf://owner/repo/filename` urls. I rewrote the parser, and the new parser will parse both `hf:ower/repo/filename` and `hf://owner/repo/filename` (as well as e.g. `huggingface:owner/repo/filename`

PR #454, opened 2026-03-26
Feat/uniffi javascript

s194042:


pielouNW:
I currently can't integrate the react-native package  from npmjs in the react-native example project because this package needs to be integrated `https://www.npmjs.com/package/uniffi-bindgen-react-native` within `react-native-nobodywho`. 

Here is the error on iOS:
```
Unable to resolve module uniffi-bindgen-react-native from /Users/pielou/Documents/GitHub/react-native-starter-example/node_modules/react-native-nobodywho/generated/ts/nobodywho.ts: uniffi-bindgen-react-native could not be found within the project or in these directories:
  node_modules
  68 |   uniffiTraitInterfaceCall,
  69 |   uniffiTypeNameSymbol,
> 70 |   variantOrdinalSymbol } from "uniffi-bindgen-react-native";
```
And on Android:
```
react-native-starter-example/node_modules/react-native-nobodywho/cpp/../generated/cpp/nobodywho.hpp:5:10: fatal error: 'UniffiCallInvoker.h' file not found
      5 | #include "UniffiCallInvoker.h"
```

Once I've hot fixed this by adding `uniffi-bindgen-react-native` with `npm install uniffi-bindgen-react-native`, I get this error (on iOS):
```
TurboModuleRegistry.getEnforcing(...): 'Nobodywho' could not be found. Verify that a module by this name is registered in the native binary.
```

Here is a potential [fix](https://github.com/nobodywho-ooo/nobodywho/commit/7317ed1ca14d80e354b0d3ce0d34a9f1f2b382e5) for this two errors.

docs/markdown/react-native/chat.md
```diff
@@ -0,0 +1,203 @@
...
+```typescript
+import { Chat } from "react-native-nobodywho";
+
+const chat = await Chat.fromPath({
+  modelPath: "/path/to/model.gguf",
+  systemPrompt: "You are a mischievous assistant!",
+});
+```
+
+This `systemPrompt` is then persisted until the chat context is reset.
```
AsbjornOlling, line 97: I think the reset history function actually keeps the system prompt

docs/markdown/react-native/chat.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Context
+
+The context is the text window which the LLM currently considers. Specifically this is the number of tokens the LLM keeps in memory for your current conversation.
+A bigger context size means more computational overhead, so it makes sense to constrain it. This can be done with the `contextSize` setting at creation time:
+
+```typescript
+const chat = await Chat.fromPath({
+  modelPath: "/path/to/model.gguf",
+  contextSize: 4096,
```
AsbjornOlling, line 107: Is this parameter named differently in different places?

docs/markdown/react-native/chat.md
```diff
@@ -0,0 +1,203 @@
...
+Chat templates are used internally by models to format conversation history into the expected prompt format. Different models may support different template variables that control specific behaviors. Template variables are boolean flags passed to the chat template that can enable or disable certain features.
+
+### Using Template Variables
+
+You can set template variables when creating a chat or modify them on existing instances:
+
+```typescript
+const chat = await Chat.fromPath({
+  modelPath: "/path/to/model.gguf",
+  templateVariables: new Map([["enable_thinking", true]]),
```
AsbjornOlling, line 167: object here?

docs/markdown/react-native/embeddings-and-rag.md
```diff
@@ -0,0 +1,192 @@
...
+import { Encoder } from "react-native-nobodywho";
+
+const encoder = await Encoder.fromPath({
+  modelPath: "/path/to/embedding-model.gguf",
+});
+const embedding = await encoder.encode("What is the weather like?");
+console.log(`Vector with ${embedding.length} dimensions`);
+```
+
+The resulting embedding is an array of numbers (typically 384 or 768 dimensions depending on the model).
```
AsbjornOlling, line 32: is it always this dimensionality? Might not be relevant

docs/markdown/react-native/embeddings-and-rag.md
```diff
@@ -0,0 +1,192 @@
...
+```
+Query: "What are the office hours for customer support?"
+Documents: [
+    "Customer asked: What are the office hours for customer support?",
+    "Support team responds: Our customer support is available Monday-Friday 9am-5pm EST",
+    "Note: Weekend support is not available at this time"
+]
+```
+
+Using embeddings alone, the first document scores highest (most similar to the query) even though it provides no useful information. A cross-encoder correctly identifies that the second document actually answers the question.
```
AsbjornOlling, line 81: This runs counter to my intuition. I'd expect  the cross-encoder to also rank the first one highest. Is this true?
Maybe test, or omit

docs/markdown/react-native/tool-calling.md
```diff
@@ -0,0 +1,109 @@
...
+sidebar_title: Tool Calling
+order: 2
+---
+
+To give your LLM the ability to interact with the outside world, you will need tool calling.
+
+!!! info ""
+    Note that **not every model** supports tool calling. If the model does not have
+    such an option, it might not call your tools.
+    For reliable tool calling, we recommend trying the [Qwen](https://huggingface.co/Qwen/models) family of models.
```
AsbjornOlling: We could link to the NW huggingface collection of qwen3 models here.

docs/markdown/react-native/tool-calling.md
```diff
@@ -0,0 +1,109 @@
...
+function getCurrentDir(): string {
+  return "/home/user/documents";
+}
+
+function listFiles(args: Record<string, unknown>): string {
+  // In a real app, you'd read the filesystem here
+  return "Files: report.pdf, notes.txt, model.gguf";
+}
+
+function getFileSize(args: Record<string, unknown>): string {
```
AsbjornOlling: Maybe it would be cleaner to implement these functions in-line, like we do with the previous examples.

docs/markdown/react-native/tool-calling.md
```diff
@@ -0,0 +1,109 @@
...
+
+const response = await chat
+  .ask("What is the biggest file in my current directory?")
+  .completed();
+console.log(response);
+```
+
+## Tool calling and the context
+
+As with everything made to improve response quality, using tool calls fills up the context faster than simply chatting with an LLM. So be aware that you might need to use a larger context size than expected when using tools.
```
AsbjornOlling: 🤔 nitpick but: not *everything* made to improve repsonse quality uses up more context. e.g. samplers don't.

docs/markdown/react-native/vision.md
```diff
@@ -0,0 +1,102 @@
...
+
+```typescript
+const chat = await Chat.fromPath({
+  modelPath: "/path/to/vision-model.gguf",
+  projectionModelPath: "/path/to/mmproj.gguf",
+  contextSize: 8192,
+});
+```
+
+Or, for example, preprocess your images with some kind of compression (sometimes even changing the image type helps).
```
AsbjornOlling: Does compression help? I would expect that e.g. applying jpeg compression still results in a bitmap of the same size. I think downscaling is what we need to recommend.

nobodywho/core/src/chat.rs
```diff
@@ -65,8 +65,8 @@ pub enum Role {
 
 #[derive(Deserialize, Serialize, Clone, PartialEq, Eq, Debug, Hash)]
 pub struct Asset {
-    id: String,
-    path: PathBuf,
+    pub id: String,
+    pub path: PathBuf,
```
AsbjornOlling, line 69: why?

nobodywho/react-native/android/src/main/java/com/nobodywho/NobodywhoModule.kt
```diff
@@ -0,0 +1,44 @@
+// Generated by uniffi-bindgen-react-native
+package com.nobodywho
```
AsbjornOlling: we don't have nobodywho.com. This should be ooo.nobodywho or ai.nobodywho

nobodywho/react-native/android/build.gradle
```diff
@@ -0,0 +1,239 @@
...
+          e
+        )
+      }
+    }
+  }
+}
+
+android {
+  if (supportsNamespace()) {
+    namespace "com.nobodywho"
```
AsbjornOlling: we don't have nobodywho.com

nobodywho/react-native/DEVELOPMENT.md
```diff
@@ -0,0 +1,246 @@
...
+
+### Testing on Android
+
+#### Build the test app
+
+```bash
+# From project root
+nix develop .#android --command bash -c \
+  'cd nobodywho/react-native/test-app/android && \
+   ANDROID_HOME=$ANDROID_SDK_ROOT \
```
AsbjornOlling: might be cleaner to set ANDROID_HOME in android.nix

docs/markdown/react-native/chat.md
s194042: Reset history does, but reset context does not. We have a naming inconsistency/

docs/markdown/react-native/chat.md
s194042: Yes. but I think it should be renamed in python.

docs/markdown/react-native/embeddings-and-rag.md
s194042: It does say typically. I think it is fine.

nobodywho/core/src/chat.rs
s194042: Because uniffi is a bit ugly, so we cannot simply mirror types like we do in flutter rust bridge but have to create new message, asset etc types, which we convert into the NW types when needed. This means that we need to access these fields on Asset, because we have to go both ways.

nobodywho/react-native/DEVELOPMENT.md
s194042: Agreed.

PR #480, opened 2026-04-17
Fix/linux pytest

s194042:


AsbjornOlling:
@s194042 I also applied a fix for the linting error on #423 
See commit 4cba2d1

PR #481, opened 2026-04-17
feat: qwen3.5 & 3.6 tool calling

krebso:
## Description
Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

## Checklist:


AsbjornOlling:
The linting error from CI is fixed on main, if you rebase it, tests should pass.

.github/workflows/python_ci.yml
```diff
@@ -394,7 +398,7 @@ jobs:
           TEST_MODEL: ${{ github.workspace }}/models/${{ matrix.model.file }}
         run: |
           source .venv/bin/activate
-          python3 -m pytest tests/test_tool_calling.py -v --tb=short
+          python3 -m pytest ${{ matrix.model.pytest_args }}
```
AsbjornOlling: Am I missing something about this abstraction? Looks like all of the `pytest_args` fields are identical.

nobodywho/core/src/tool_calling/qwen35_36.rs
```diff
@@ -0,0 +1,361 @@
...
+impl ToolFormatHandler for Qwen35_36Handler {
+    fn begin_token(&self) -> &str {
+        BEGIN_TOKEN
+    }
+
+    fn end_token(&self) -> &str {
+        END_TOKEN
+    }
+
+    fn generate_grammar(&self, tools: &[Tool]) -> Result<GbnfGrammar, ToolFormatError> {
```
AsbjornOlling, line 34: It seems like this function generates invalid grammars. It fails the tests, getting a null result from initializing the grammar.

On e.g. the bash tool test, I get this grammar:
```
qwen35-val-body ::= [^]*
qwen35-t0-p-commands-block ::= "<parameter=commands>\n" qwen35-val-body "\n</parameter>\n"
qwen35-t0-call ::= "<function=run_bash>\n" qwen35-t0-p-commands-block "</function>\n"
tool-alt ::= qwen35-t0-call
toolcall ::= "<tool_call>" "\n" tool-alt "</tool_call>"
superroot ::= toolcall+
```

My guess is that empty negative character sets are not allowed. I also think that it doesn't make much sense in general? We might want to disallow termination tokens, and maybe also newlines?

.github/workflows/python_ci.yml
```diff
@@ -345,18 +345,22 @@ jobs:
...
             file: functiongemma-270m-it-BF16.gguf
             url: https://huggingface.co/unsloth/functiongemma-270m-it-GGUF/resolve/main/functiongemma-270m-it-BF16.gguf
             format: FunctionGemma
+            pytest_args: tests/test_tool_calling.py -v --tb=short
           - name: Ministral3
             file: mistralai_Ministral-3-3B-Instruct-2512-Q6_K_L.gguf
             url: https://huggingface.co/bartowski/mistralai_Ministral-3-3B-Instruct-2512-GGUF/resolve/main/mistralai_Ministral-3-3B-Instruct-2512-Q6_K_L.gguf
             format: Ministral3
+            pytest_args: tests/test_tool_calling.py -v --tb=short
           - name: Gemma4
```
AsbjornOlling, line 360: We could add the smallest qwen3.5 model here, to make sure that it works with the tool calling format.

Sure, the 30B model is too big for GH Actions, but 0.8B is probably not?

AsbjornOlling (changes requested):

AsbjornOlling (approved):

PR #482, opened 2026-04-17
feat: rust crate implementation

TheGoldenChicken:
## Description

First version of a rust crate implementation, by making an interface that follows convention in the Python module. Most code in rust/src/lib.rs was generated using Claude. Rust interface uses builder pattern.

Often was difficult to know when something made sense to make an interface for, and when it easier to just refer to the existing rust implementations - there is a discussion to be had on many things there, including some minor ones, such as naming.

One **bigg(tm)** issue is that crates.io requires all dependencies to be on crates.io already. We currently have **gbnf**, **monty** **bashkit**, and the current version of **llama-cpp**, which are not published. Llama will fix itself, obviously, but monty are currently waiting for stuff to be published upstream themselves, and gbnf will obviously need to be published by ourselves.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?
Ran integration test suite based on tests in the Python module

## Checklist:
- [ ] My changes generate no new warnings (unsure about some warnings generated)


TheGoldenChicken:
Right now, appears there is no "smart" way to get around the requirement - that all dependencies have to be published to crates.io. Llama-cpp will likely work itself out, and bashkit might not be necessary, gbnf we can publish outselves. [Monty, however, is still waiting upstream for Ruff](https://github.com/pydantic/monty/issues/148#issuecomment-3885572032), so no dice. 

Current solution is to have users install through Github, and then just have them install dependencies manually. We **may** be able to help them there, by posting some kind of semi-stub nobodywho package to crates.io that essentially just installs nobodywho proper, though that is kind of a malware-y solution. In any case, **we should most definitely squat on the crates.io name "nobodywho"**, to prevent uesrs from installing junk programs when they thought they could cargo add nobodywho. This can also work to enhance discoverability of the package.

AsbjornOlling:
stale and unfinished. closing

PR #498, opened 2026-04-25
on_download_progress callback

AsbjornOlling:
Callbacks to show progress bar when downloading YEAH

I added it to all of the bindings:

### Flutter

Because `Option<...callable thingy...>` is not supported by FRB, this defaults to a no-op callback (not the indicatif one). Otherwise you can pass a callback that takes two ints (downloaded so far and total). This one blocks on async functions, so if the async function is slow, it may slow down the download.

### Python

passing `None` to the progress callback param defaults to an indicatif progess bar. otherwise the regular callback is invoked with two ints

### React Native

A regular js callback is passed in, called with two ints. If an async function is called, it is not awaited (fire-and-forget). Defaults to the indicatif progress bar.

### Godot

This one is a bit trickier, since you can only emit signals on the main thread. I pass in a callback that has a queue on it, and then call select in a loop to emit the stuff on a signal.
Godot is signal-based, not callbacks.

### DIFF
--- a/docs/markdown/python/index.md
+++ b/docs/markdown/python/index.md
@@ -35,4 +35,17 @@ print(full_response)
 
 This is a super simple example, but we believe that examples which do simple things, should be simple!
 
+## Tracking download progress
+
+When loading a remote model, pass an `on_download_progress` callback to observe the download. It receives `(downloaded_bytes, total_bytes)` and is not called for cached or local files. If you don't pass anything, NobodyWho prints a default terminal progress bar.
+
+```python
+from nobodywho import Model
+
+model = Model(
+    'huggingface:NobodyWho/Qwen_Qwen3-0.6B-GGUF/Qwen_Qwen3-0.6B-Q4_K_M.gguf',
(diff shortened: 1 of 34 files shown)


docs/markdown/python/index.md
```diff
@@ -35,4 +35,17 @@ print(full_response)
...
+## Tracking download progress
+
+When loading a remote model, pass an `on_download_progress` callback to observe the download. It receives `(downloaded_bytes, total_bytes)` and is not called for cached or local files. If you don't pass anything, NobodyWho prints a default terminal progress bar.
+
+```python
+from nobodywho import Model
+
+model = Model(
+    'huggingface:NobodyWho/Qwen_Qwen3-0.6B-GGUF/Qwen_Qwen3-0.6B-Q4_K_M.gguf',
+    on_download_progress=lambda d, t: print(f"{d}/{t} bytes"),
```
marek-hradil:
```suggestion
    on_download_progress=lambda downloaded, total: print(f"{downloaded}/{total} bytes"),
```

marek-hradil (approved):

Issue #501, opened 2026-04-27
[FEATURE] Function to list cached models


TheGoldenChicken:
Says on the tin, some function exposed that allows users to list all models that have been downloaded and cached by nobodywho. Useful in cases where an application allows users to switch between different models - which will lead to the cache just bloating over time. Probably would be too specific to make a "maximum_cache_size" option - this at least gives developers a straightforward way to control that themselves.

Likely another similarly convenient thing would be listing the current size of all models cached by nobodywho. Kinda pedantic, maybe, but convenient.

AsbjornOlling:
Closed with #508

PR #508, opened 2026-04-29
get_cached_models

TheGoldenChicken:
Addresses #501 

Adds a small function to core, exposed in integrations, which lists the path of all currently cached models (those downloaded by Nobodywho), as well as the size they take in bytes. 

People *can* probably get these without our help... this is just a somewhat more convenient solution for them... that way, they don't have to guess out where exactly nobodywho caches their models.

This PR also includes some re-structuring of the errors emitted during model downloading, since the get_cached_models function now also calls the get_cache_dir function.


AsbjornOlling:
rebased on main

s194042 (approved):

PR #512, opened 2026-05-01
run mac CI tasks on warp

AsbjornOlling:


AsbjornOlling:
This ran really fast, which is nice - *but* it also spent $6 immediately for a single run :/ Might be super overkill

s194042 (approved):

PR #516, opened 2026-05-06
feat: Llama-3.x + LFM2.5 family tool calling

### LINKED ISSUE #214: [FEATURE] Tool calling GBNF grammar for Llama tool call format
The llama models are famous and widespread. It would be cool to support grammar-constrained generation for their tool calling format too.

### LINKED ISSUE #440: [FEATURE] Enable logs on Flutter

Image

ahmedibrahim085:
## Description

Adds two new model-family tool-calling handlers and three cross-cutting fixes that benefit every existing handler.

**Closes #214** ("[FEATURE] Tool calling GBNF grammar for Llama tool call format"). Supersedes the closed-without-review #326 with a shape-correct handler matching the merged Gemma4 (#464) and Qwen3.5/3.6 (#481) precedent (separate `core/src/tool_calling/<family>.rs` module + GBNF grammar generation + unit tests).

## Type of change

- [x] New feature (Llama-3.x handler + LFM2.5 family handler — non-breaking, additive enum variants)
- [x] Bug fix (3 cross-cutting fixes that affect ALL existing handlers — net-positive correctness)

### New family handlers
- `feat(tool_calling): add Llama-3.x format handler` — `core/src/tool_calling/llama32.rs` + 16 unit tests. Format: `<|python_tag|>{"name"|"function": "...", "parameters": {...}}<|eot_id|>` (prefix optional in extraction; model often omits it).
- `feat(tool_calling): wire Llama32 handler into detect_tool_format` — template-marker (`Environment: ipython` / `<|python_tag|>`) + arch + name fallback chain in `mod.rs`.
- `feat(tool_calling): LFM2.5 family handler in parser-only mode` — `core/src/tool_calling/lfm2_5.rs` + 17 unit tests. Format: `<|tool_call_start|>[fn1(arg=val), fn2(arg=val)]<|tool_call_end|>` Pythonic call list. Handler returns `Err` from `generate_grammar` so the engine path keeps `tool_format=Some(handler)` but `tool_grammar=None`; sampler runs unconstrained, parser still extracts on output. Mirrors llama.cpp's pre-autoparser working configuration documented in [llama.cpp #20245](https://github.com/ggml-org/llama.cpp/issues/20245).
- `feat(tool_calling): detect LFM2.5 family by chat-template marker` — adds `template_str.contains(\"keep_past_thinking\")` template-marker check before the metadata fallback. Necessary because LFM2.5-1.2B-Instruct and LFM2.5-1.2B-Thinking GGUFs ship with `general.name` set to a git-commit SHA (e.g. `4cd563d5...`), bypassing the name-substring fallback.
- `feat(tool_calling/lfm2_5): handle Pythonic list/set/dict/tuple literals` — extends parser to recognise `[...]`/`(...)`/`{...}` value shapes that small models emit when the user prompt mirrors Python literal syntax. Adds `has_top_level_colon` helper to discriminate set vs dict literals using a bracket/quote-aware scanner.

### Cross-cutting fixes (apply to ALL handlers)
- `fix(tool_calling): preserve OpenAI key order in Tool serialization` — rewrites `Serialize for Tool` using nested `serialize_map` with explicit `serialize_entry` calls (outer `type` then `function`; inner `name` → `description` → `parameters`). The previous `serde_json::json!({...}).serialize()` path materialised a `Value::Object` backed by `BTreeMap`, which re-sorted keys alphabetically. LFM2.5-350M is empirically sensitive to that order: alphabetical (`{\"function\":..., \"type\":\"function\"}`) makes it hallucinate extra Pythonic kwargs (`parameters=`, `type=`), breaking the parser. Avoids forcing the `serde_json/preserve_order` cargo feature globally.
- `feat(template): pass tools to chat template as pre-serialized JSON strings` — pre-serialise each `Tool` to a JSON STRING via `serde_json::to_string(t)` and pass `Vec<String>` as the `tools` template var. The HF tool-aware chat templates (LFM2/2.5/Qwen3 etc.) all guard with `{%- if tool is not string -%}{%- set tool = tool | tojson -%}{%- endif -%}` so a pre-serialised string passes through unchanged. This bypasses minijinja's internal `Value::from_serialize` round-trip through `BTreeMap`.
- `fix(flutter): drop additionalProperties:false from generated tool schemas` — `dart_function_type_to_json_schema` no longer injects `additionalProperties: false` into the JSON schema returned for each Dart tool function. Five `mod tests` expected-schema fixtures updated to drop the same field. Required because small models echo `additionalProperties` as a literal Pythonic kwarg in the emitted tool call (e.g. `[circle_area(additionalProperties={\"radius\": 5}, ...)]`), breaking the parser. The field is OpenAI-strict-mode shorthand that is NOT required by the llama.cpp parser path.

### Engine reliability fix
- `fix(chat): emit terminal Done after tool-call loop epilogue` — adds an unconditional `respond(WriteOutput::Done(response))` at the end of `Worker::ask()`'s chat-loop. Required because `wrap_respond` flips `emitting=false` once the tool-call begin token is observed, suppressing all subsequent forwards including the terminal `Done`. When `extract_tool_calls` returns `None` on a parser-only path (model emitted markers but the parser couldn't recognise the inner Pythonic), the chat-loop exited without ever signalling completion to the consumer; `TokenStream::completed` saw a closed channel with `completed_response = None` and surfaced that as `WorkerCrashed`. `TokenStream::next_token` short-circuits on `completed_response.is_some()`, so multi-iteration flows where iter-N already delivered a Done (Qwen3 etc.) simply ignore this terminal one — no double-process, no regression.

### Host
- `cargo test --release --lib tool_calling`: **66 passed / 0 failed / 2 ignored** (was 53 before this PR; +13 new tests).
- `cargo test --release --lib --workspace`: same 19 fixture-only failures as upstream main (`ModelNotFound` integration tests on `model.gguf` / `embeddings.gguf` / `crossencoder.gguf`); no regressions.
- Existing handler test counts preserved (Qwen3, Gemma4, FunctionGemma, Ministral3, Qwen3.5/3.6 all green).

### Device (Samsung Galaxy S22)
Round r21 bake-off via `bake_off_lfm2_5_family.sh` (5 prompts × 3 sizes), with internet connectivity and screen-on:

| model | p1 (circle r=5) | p2 (circle r=12.5) | p3 (Cairo) | p4 (London) | p5 (Tokyo) | reach | crashes |
|---|---|---|---|---|---|---|---|
| LFM2.5-350M | NUMERIC_OK | NUMERIC_PRESENT | TOOL_OK | TOOL_OK | TOOL_OK | 5/5 | 0 |
| LFM2.5-1.2B-Instruct | NUMERIC_OK | NUMERIC_OK | TOOL_OK | TOOL_OK | TOOL_OK | 5/5 | 0 |
| LFM2.5-1.2B-Thinking | NUMERIC_OK | NUMERIC_OK | DNF (thinking-mode latency >maestro budget) | DNF | DNF | 2/5 reach | 0 |

p2 NUMERIC_PRESENT on 350M is small-model arithmetic on 12.5²×π — pure model limit, not a handler bug.

### Cross-validation
Reference rendered prompt captured from `llama-server --jinja` for LFM2.5-350M; engine's rendered prompt now matches byte-for-byte after the cross-cutting fixes (key order + tools-as-strings).

## Known gaps NOT addressed in this PR

1. **`python_ci.yml` matrix entry deferred for LFM2.5**. Local-Mac validation reveals an upstream `ggml-metal` cleanup assertion (related to issue [ggml-org/llama.cpp#22593](https://github.com/ggml-org/llama.cpp/issues/22593) and PR [#22595](https://github.com/ggml-org/llama.cpp/pull/22595), but the marek-hradil fork has a different organisation around `ggml_metal_buffer_free` that needs a separate fix likely in the Rust Drop chain). On non-Metal Linux CI this should not surface; maintainer welcome to add the matrix entry after CI dry-run. Llama-3.x matrix entry (`Llama32` format) not added pending maintainer preference on which Llama-3.2-1B GGUF to download.
2. **`Cargo.nix` regen pending**. New `tracing-android` deps live on a sibling branch that closes [#440](https://github.com/nobodywho-ooo/nobodywho/issues/440); separate PR forthcoming. This branch keeps `Cargo.lock` and `Cargo.nix` in sync with origin/main.
3. **LFM2.5 model capability on heavy multi-tool fixtures with set/tuple/dict types**. Distil labs benchmark reports 34-63% baseline on 350M for those workloads — pure model capability, no engine fix possible. Not blocked by this PR; tools fire correctly when model decides to use them.

## Checklist:

- [x] My code follows the style guidelines of this project (matched #464 / #481 file shape: new handler module + mod.rs wiring + tests)
- [x] I have added tests that prove my fix is effective or that my feature works (38 new unit tests across llama32.rs and lfm2_5.rs + 5 cross-cutting test fixture updates)

## Notes for reviewers

- 9 commits with cleanup history preserved (same cadence as #481's 16 commits and #464's 14 commits).
- Cross-cutting fixes (key order, additionalProperties drop, tools-as-strings) are net-positive for all existing handlers — they were prerequisites for LFM2.5 to parse cleanly but the underlying bugs (BTreeMap key re-sort + minijinja round-trip) affected every family's prompt rendering. Verified Qwen3 + Llama-3.x + Gemma4 + FunctionGemma + Ministral3 still pass their unit tests.

### DIFF
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -2634,9 +2634,14 @@ mod tests {
         test_utils::init_test_tracing();
         let model = test_utils::load_test_model();
 
-        // Use a very small context size to force shifting
+        // Use a very small context size to force shifting. Bumped n_messages
+        // from 10 -> 20 because passing tools to the chat template as
+        // pre-serialized JSON strings (see commit a26dfc79) reduced the
+        // per-render token count enough that 10 exchanges no longer overflow
+        // n_ctx=1024 with the test_chat-model GGUF — context_shift then
+        // skipped, leaving messages_after.len() == messages_before.
         let n_ctx = 1024;
-        let n_messages = 10;
+        let n_messages = 20;
         let mut worker = Worker::new_chat_worker(
             &model,
             ChatConfig {
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -50,7 +50,7 @@ use std::cmp::min;
 use std::collections::HashSet;
 use std::hash::Hasher;
 use std::path::PathBuf;
-use std::sync::atomic::AtomicBool;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, MutexGuard};
 use tracing::{debug, error, info, trace, trace_span};
 
@@ -1643,11 +1643,21 @@ impl Worker<'_, ChatWorker> {
             },
         );
 
+        // Tracks whether any iteration's Done event was forwarded to the outer
+        // callback. Shared across every `wrap_respond` instance built below so
+        // the terminal-Done epilogue can run only when no Done has reached the
+        // consumer yet (e.g. parser-only paths that exit before producing an
+        // iter-2 generation). Keeps raw `Fn(WriteOutput)` consumers — godot
+        // binding, direct ask_channel users — from receiving two Done events
+        // per `ask` call.
+        let done_forwarded = Arc::new(AtomicBool::new(false));
+
         // get the finished response
         let mut response: String = self.wrapped_update_context_and_generate_response(
             sampler.clone(),
             respond.clone(),
             tool_call_begin.clone(),
+            done_forwarded.clone(),
         )?;
 
         // Process tool calls if tool format is configured
@@ -1691,6 +1701,7 @@ impl Worker<'_, ChatWorker> {
                     sampler.clone(),
                     respond.clone(),
                     tool_call_begin.clone(),
+                    done_forwarded.clone(),
                 )?;
             }
         } // Close if let Some(tool_format)
@@ -1702,18 +1713,18 @@ impl Worker<'_, ChatWorker> {
 
         self.extra.context.chunks = self.render_as_chunks(true)?;
 
-        // Terminal Done. Guarantees the consumer's TokenStream::completed
-        // (or any Fn(WriteOutput) callback) ALWAYS receives a final Done
-        // event, even when the chat-loop exits early because
-        // `extract_tool_calls` returned None on a parser-only path that
-        // never produced an iter-2 generation. `wrap_respond` suppresses
-        // Done after the begin_token in iter 1, which used to surface as
-        // `WorkerCrashed` for LFM2.5 and other parser-only handlers.
-        // `TokenStream::next_token` short-circuits when
-        // `completed_response.is_some()`, so a multi-iteration flow
-        // (Qwen3 etc.) that already received its iter-N Done simply
-        // ignores this one — no double-process.
-        respond(llm::WriteOutput::Done(response));
+        // Terminal Done failsafe. Fires only when no iteration forwarded a Done
+        // — i.e. parser-only paths where `extract_tool_calls` returned None
+        // after `wrap_respond` already suppressed iter-1's Done past the
+        // begin_token. That used to surface as `WorkerCrashed` for LFM2.5 and
+        // other parser-only handlers. Skipping this when a Done was already
+        // forwarded prevents a double Done for raw `Fn(WriteOutput)` consumers
+        // (godot binding, direct ask_channel users); `TokenStream` consumers
+        // were already shielded by the `completed_response.is_some()`
+        // short-circuit, but external callback users were not.
+        if !done_forwarded.load(Ordering::SeqCst) {
+            respond(llm::WriteOutput::Done(response));
+        }
 
         Ok(self)
     }
@@ -1757,6 +1768,7 @@ impl Worker<'_, ChatWorker> {
         sampler: SamplerConfig,
         respond: F,
         tool_call_begin_token: Option<String>,
+        done_forwarded: Arc<AtomicBool>,
     ) -> Result<String, WrappedResponseError>
     where
         F: Fn(llm::WriteOutput) + Clone,
@@ -1768,7 +1780,8 @@ impl Worker<'_, ChatWorker> {
 
         // wrap the response callback to keep a copy of the completed response
         // and to avoid emitting tool calls
-        let (wrapped_respond, resp_receiver) = wrap_respond(respond.clone(), tool_call_begin_token);
+        let (wrapped_respond, resp_receiver) =
+            wrap_respond(respond.clone(), tool_call_begin_token, done_forwarded);
 
         // llm go brrr
         self.generate_response_until_done(sampler, wrapped_respond, &inference_lock_token)?;
@@ -1965,12 +1978,15 @@ impl Worker<'_, ChatWorker> {
     }
 }
 
-/// wraps a response function in a closure to do two things:
+/// wraps a response function in a closure to do three things:
 /// 1. save a copy of the response (using a channel) before sending it out
 /// 2. skip emitting once a tool_call_begin_token has been seen
+/// 3. record whether a Done was forwarded to the outer callback (so the caller
+///    can decide whether a terminal-Done epilogue is still needed)
 fn wrap_respond<F>(
     respond: F,
     tool_call_begin_token: Option<String>,
+    done_forwarded: Arc<AtomicBool>,
 ) -> (
     impl FnMut(llm::WriteOutput),
     std::sync::mpsc::Receiver<String>,
@@ -1990,6 +2006,9 @@ where
                 resp_sender
                     .send(resp.clone())
                     .expect("Failed sending response");
+                if emitting {
+                    done_forwarded.store(true, Ordering::SeqCst);
+                }
             }
             llm::WriteOutput::Token(_) => (),
         }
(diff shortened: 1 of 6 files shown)


nobodywho/core/src/chat.rs
```diff
@@ -1698,10 +1698,23 @@ impl Worker<'_, ChatWorker> {
...
+        // (or any Fn(WriteOutput) callback) ALWAYS receives a final Done
+        // event, even when the chat-loop exits early because
+        // `extract_tool_calls` returned None on a parser-only path that
+        // never produced an iter-2 generation. `wrap_respond` suppresses
+        // Done after the begin_token in iter 1, which used to surface as
+        // `WorkerCrashed` for LFM2.5 and other parser-only handlers.
+        // `TokenStream::next_token` short-circuits when
+        // `completed_response.is_some()`, so a multi-iteration flow
+        // (Qwen3 etc.) that already received its iter-N Done simply
+        // ignores this one — no double-process.
```
gergelyvagujhelyi: Are you sure?

nobodywho/core/src/chat.rs
```diff
@@ -1698,10 +1698,23 @@ impl Worker<'_, ChatWorker> {
...
+        respond(llm::WriteOutput::Done(response));
```
gergelyvagujhelyi: #[test]
    fn test_done_emitted_exactly_once_per_ask() -> Result<(), Box<dyn std::error::Error>> {
        let model = test_utils::load_test_model();

        let mut worker = Worker::new_chat_worker(
            &model,
            ChatConfig {
                n_ctx: 1024,
                ..Default::default()
            },
            Arc::new(AtomicBool::new(false)),
        )?;

        let done_count = Arc::new(std::sync::atomic::AtomicUsize::new(0));
        let count_inner = Arc::clone(&done_count);
        let (sender, receiver) = std::sync::mpsc::channel();

        let callback = move |x: llm::WriteOutput| {
            if let llm::WriteOutput::Done(resp) = x {
                count_inner.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
                // tolerate a closed channel — count is the assertion target,
                // recv-side just unblocks once.
                let _ = sender.send(resp);
            }
        };

        worker.ask("Say hi.".into(), callback)?;

        // First Done: confirm a callback arrived (proves wiring).
        let _ = receiver.recv_timeout(std::time::Duration::from_secs(60))?;

        // `Worker::ask` returns synchronously after invoking its callback for
        // the final time, so by here the count is final — no race window.
        let count = done_count.load(std::sync::atomic::Ordering::SeqCst);
        assert_eq!(
            count, 1,
            "Worker::ask should emit exactly ONE Done per call, observed {count}. \
             A second Done from the unconditional `respond(Done)` after the tool \
             loop (chat.rs:1716) double-fires for raw-callback consumers (godot \
             binding, direct ask_channel users). TokenStream consumers are \
             shielded by the `completed_response.is_some()` short-circuit but \
             external users of `Fn(WriteOutput)` callbacks are not."
        );

        Ok(())
    }

AsbjornOlling:
Thanks for submitting this! It looks pretty reasonable overall. Love to see it.

But this is a lot of code in one PR, tackling a handful of different things, making it difficult to review. Any chance you could split this into a few independent PRs instead?

Also, if you could commit it in a such a way that claude is not listed as a contributor, that would be good.

ahmedibrahim085:
Thanks @gergelyvagujhelyi for the review and the test. You're right — the comment in the original commit was wrong on the broader picture: `TokenStream` consumers were shielded by `completed_response.is_some()`, but raw `Fn(WriteOutput)` callback consumers (godot binding, direct `ask_channel` users) saw two Done events whenever an iteration's Done already reached the outer callback.

Fixed in e9e16266: `wrap_respond` now also tracks whether it forwarded a Done via a shared `Arc<AtomicBool>`, and the terminal-Done epilogue fires only when no iteration forwarded one. Preserves the LFM2.5 / parser-only failsafe (where iter-1's Done is suppressed by `wrap_respond` and `extract_tool_calls` may return None on iter-2) without double-firing on grammar paths.

Your test is added (lightly adapted to match the existing `test_chat_worker` style) as `chat::tests::test_done_emitted_exactly_once_per_ask` — passes locally with an LFM2 model.

@AsbjornOlling — re: split + claude trailer scrub, working on both. Will report back once the branches are reorganised.

ahmedibrahim085:
Superseded by 3 smaller PRs, per @AsbjornOlling's review:

- #519 — engine fixes (key order, tools-as-strings, drop additionalProperties, test tuning) — affects all families
- #520 — Llama-3.x family handler — independent of #519
- #521 — LFM2.5 family handler (parser-only) + terminal Done epilogue — stacked on #519

The double-Done bug @gergelyvagujhelyi found is fixed in #521 with the regression test from review (`test_done_emitted_exactly_once_per_ask`).

All commits on the new branches are cleaned of the `Co-Authored-By: Claude` trailer, per @AsbjornOlling's request.

Closing this PR — please continue review on the splits.

PR #517, opened 2026-05-06
feat(flutter): route engine tracing through Android logcat

### LINKED ISSUE #440: [FEATURE] Enable logs on Flutter

ahmedibrahim085:
## Description

Adds Android-only deps under `[target.'cfg(target_os = \"android\")'.dependencies]` and an `init_android_logging` helper invoked from `init_app` when the target is Android. Engine `tracing::info!` / `debug!` / `warn!` events now reach `adb logcat -s nobodywho`.

**Closes #440** ([FEATURE] Enable logs on Flutter).

## Type of change

- [x] New feature (Android-only diagnostics; non-breaking on iOS / web / desktop)

## Background

Previously `flutter_rust_bridge::setup_default_user_utils()` + `tracing_subscriber::fmt()` wrote to stderr, which on Android release builds is silently discarded. Engine `debug!()` / `error!()` events from `core/src/chat.rs`, `tool_calling/*.rs`, `template.rs`, etc. never reached logcat — making any tool-calling regression on device invisible without grepping the model output text itself.

The uniffi crate at `nobodywho/uniffi/src/lib.rs:9-26` already uses the same pattern (`android_logger 0.14`); this PR brings the flutter binding to parity.

## What's in the PR (3 commits)

- `feat(flutter): wire tracing-android layer for logcat output` — `flutter/rust/Cargo.toml` adds `android_logger`, `tracing-log`, `log`, `tracing-android` (Android-only target deps); `flutter/rust/src/lib.rs::init_app` installs `tracing_subscriber::registry().with(tracing_android::layer(\"nobodywho\"))` on Android with DEBUG max level. Plain `tracing_subscriber::fmt()` path retained for non-Android targets behind `#[cfg(not(target_os = \"android\"))]`.
- `feat(chat): instrument render_as_chunks with rendered-prompt tracing` — single `info!(target: \"nobodywho_prompt\", ..., \"RENDERED PROMPT BEGIN\\n{}\\nRENDERED PROMPT END\", rendered_chat)` line inside `render_as_chunks`. Wrapped between markers so logcat slices can be extracted with `awk`. Reviewer welcome to downgrade to `debug!` if `info!` is too verbose for the default level.
- `chore(lockfile): regenerate Cargo.lock after android tracing-android deps` — reflects the new dev/prod deps added on the android target.

## How Has This Been Tested?

Filter on device with: `adb logcat -s nobodywho:* nobodywho_log:* flutter:*`. Verified working during PR #516 device evidence collection on Samsung Galaxy S22 (Android 15). Engine debug events now visible inline with Flutter `[BAKEOFF]` events.

Non-Android targets unaffected (host `cargo test` 66/0/2 unchanged).

## Checklist:

- [x] My code follows the style guidelines of this project (Android target gating mirrors `nobodywho/uniffi/src/lib.rs`)
- [x] No tests added (diagnostic-only contribution; tested via manual `adb logcat` capture)

### DIFF
--- a/nobodywho/flutter/rust/src/lib.rs
+++ b/nobodywho/flutter/rust/src/lib.rs
@@ -1011,6 +1011,30 @@ impl SamplerPresets {
     }
 }
 
+#[cfg(target_os = "android")]
+fn init_android_logging() {
+    use std::sync::Once;
+    use tracing_subscriber::layer::SubscriberExt;
+    use tracing_subscriber::util::SubscriberInitExt;
+    static INIT: Once = Once::new();
+    INIT.call_once(|| {
+        if let Ok(android_layer) = tracing_android::layer("nobodywho") {
+            tracing_subscriber::registry()
+                .with(android_layer)
+                .with(tracing_subscriber::filter::LevelFilter::DEBUG)
+                .try_init()
+                .ok();
+        }
+        // also init plain log -> android_logger for any direct log::info! callers
+        android_logger::init_once(
+            android_logger::Config::default()
+                .with_max_level(log::LevelFilter::Debug)
+                .with_tag("nobodywho_log"),
+        );
+        log::info!("NobodyWho Flutter logging initialized (tracing-android -> logcat)");
+    });
+}
+
 #[flutter_rust_bridge::frb(init)]
 pub fn init_app() {
     // send llamacpp logs into tracing
@@ -1019,18 +1043,24 @@ pub fn init_app() {
     // send logs to the appropriate places for android, ios and wasm
     flutter_rust_bridge::setup_default_user_utils();
 
-    let log_level = if cfg!(debug_assertions) {
-        tracing::Level::DEBUG
-    } else {
-        tracing::Level::INFO
-    };
+    #[cfg(target_os = "android")]
+    init_android_logging();
 
-    tracing_subscriber::fmt()
-        .with_max_level(log_level)
-        .with_timer(tracing_subscriber::fmt::time::uptime())
-        .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE)
-        .try_init()
-        .ok();
+    #[cfg(not(target_os = "android"))]
+    {
+        let log_level = if cfg!(debug_assertions) {
+            tracing::Level::DEBUG
+        } else {
+            tracing::Level::INFO
+        };
+
+        tracing_subscriber::fmt()
+            .with_max_level(log_level)
+            .with_timer(tracing_subscriber::fmt::time::uptime())
+            .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE)
+            .try_init()
+            .ok();
+    }
 }
 
 #[cfg(test)]
--- a/nobodywho/Cargo.lock
+++ b/nobodywho/Cargo.lock
@@ -59,6 +59,12 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "android_log-sys"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "85965b6739a430150bdd138e2374a98af0c3ee0d030b3bb7fc3bddff58d0102e"
+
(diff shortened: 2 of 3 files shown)


nobodywho/core/src/chat.rs
```diff
@@ -1729,6 +1729,8 @@ impl Worker<'_, ChatWorker> {
                 .render_unhandled(messages, &template_context)?
         };
 
+        info!(target: "nobodywho_prompt", handled = handled, n_msgs = messages.len(), n_tools = self.extra.tools.len(), "RENDERED PROMPT BEGIN\n{}\nRENDERED PROMPT END", rendered_chat);
```
AsbjornOlling: What is the purpose of this log line? It seems potentially very noisy, and unrelated to this PRs purpose of android log forwarding.
I think this log line should be removed.

AsbjornOlling (changes requested):

ahmedibrahim085:
Thanks for the review @AsbjornOlling. Agreed — that `info!` was leftover diagnostic from a separate LFM2.5 prompt-shape investigation and doesn't belong in this PR.

Removed in 8b27da0e. The PR now contains only the Android logcat wiring (`tracing-android` Layer + Cargo dep).

ahmedibrahim085:
Cleaned up the history: dropped the diagnostic-instrumentation commit and its follow-up removal commit (they cancelled out, leaving zero diff in `chat.rs`). PR now has 2 commits — only the load-bearing changes:

- `feat(flutter): wire tracing-android layer for logcat output` (Cargo.toml + flutter/rust/src/lib.rs)
- `chore(lockfile): regenerate Cargo.lock after android tracing-android deps`

Force-pushed to e8e3196b. Ready for re-review @AsbjornOlling.

ahmedibrahim085:
@AsbjornOlling friendly bump — diagnostic info! line removed in 8b27da0e (force-pushed to e8e3196b 2 commits, only Android logcat wiring). Ready for re-review when you have a moment.

nobodywho/flutter/rust/src/lib.rs
```diff
@@ -1011,6 +1011,30 @@ impl SamplerPresets {
...
+                .with(tracing_subscriber::filter::LevelFilter::DEBUG)
+                .try_init()
+                .ok();
+        }
+        // also init plain log -> android_logger for any direct log::info! callers
+        android_logger::init_once(
+            android_logger::Config::default()
+                .with_max_level(log::LevelFilter::Debug)
+                .with_tag("nobodywho_log"),
+        );
```
gergelyvagujhelyi, line 1033: tracing_android layer is correct, but this init_once is a no-op — verified on a Pixel 9 emulator with this PR's exact deps. setup_default_user_utils() on line 1044 already calls android_logger::init_once via FRB's default log feature, filling log 0.4.29's global set_logger slot with android_logger 0.15.1. This PR's 0.14.1 call gets SetLoggerError, silently swallowed (https://docs.rs/android_logger/0.14.1/src/android_logger/lib.rs.html); with_tag("nobodywho_log") is never registered — adb logcat -s nobodywho_log:V returns zero lines.

Fix: delete this block, drop android_logger/tracing-log/log from Cargo.toml (now unused), update commit-message filter to nobodywho:* flutter:*. tracing_android layer unaffected — bypasses log.

gergelyvagujhelyi:
Just wanted to say thanks for taking this on. Engine tracing being invisible on Android release builds has been a long-standing rough edge — #440 was open for a reason — and getting it routed to logcat genuinely makes on-device debugging less painful. The tracing_android layer was exactly the right primitive to reach for, and verifying on real hardware before posting goes a long way. We're excited about your contribution!

gergelyvagujhelyi:
Also wanted to flag — the whole team was on a business trip recently, which is why review on this came in a bit slower than usual. Definitely not a reflection of your PR. Back to normal cadence now, and thanks for your patience.

madsmtm:
I have filed https://github.com/nobodywho-ooo/nobodywho/pull/698, which instead sends logs to Dart's `logging` package. I believe this is the better approach, as it allows users to more easily configure logging.

I'll close this, but feel free to re-open if you disagree with this assessment!

PR #519, opened 2026-05-09
fix: engine fixes for tool-calling key order + template rendering

ahmedibrahim085:
Splits a chunk of #516 into smaller, independently-reviewable pieces, per @AsbjornOlling's request. This is **PR 1 of 3**.

Refs: #516, #214

## What this PR fixes

Three load-bearing tool-calling correctness fixes that benefit **all** model families (Qwen3, Llama-3.x, LFM2.5, etc.) plus a test tuning for the new tokenization. No new family handlers — those land in the follow-up PRs.

### 1. `Serialize for Tool` — preserve OpenAI canonical key order

`serde_json::json!({...})` materialises a `serde_json::Map` backed by `BTreeMap` by default, which re-sorts keys alphabetically. That made tool schemas serialize as `{"function":..., "type":"function"}` instead of the canonical `{"type":"function", "function":...}`.

Replaced with explicit `serializer.serialize_map` + `serialize_entry` calls in canonical order:
- Outer: `type` → `function`
- Inner: `name` → `description` → `parameters`

Empirically, smaller models (LFM2.5-350M in particular) are order-sensitive — alphabetical key order makes them hallucinate Pythonic kwargs in tool calls.

### 2. Pass tools to chat template as pre-serialized JSON strings

Even with a correctly ordered `Tool`, minijinja's `Value::from_serialize` round-trips through `serde_json::Value` (BTreeMap) and re-sorts keys.

Pre-serializing each `Tool` to a JSON `String` and passing `Vec<String>` as the `tools` template variable bypasses minijinja's serializer. The HF tool-aware chat templates (LFM2/2.5, Qwen3) all guard with `{%- if tool is not string -%}{%- set tool = tool | tojson -%}{%- endif -%}`, so a string passes through unchanged.

Pitfall avoided: `Option::iter()` yields the entire `Vec` (not per-Tool), producing one giant array-string. Used `.as_ref().into_iter().flatten()` to map per-Tool.

### 3. Drop `additionalProperties: false` from generated tool schemas

`flutter/rust/src/lib.rs::dart_function_type_to_json_schema` was injecting `additionalProperties: false` into the generated schema. Small models like LFM2.5-350M echoed that field name as a Pythonic kwarg in tool emits.

This is JSON Schema strict-mode shorthand from OpenAI's API spec — not required by the llama.cpp parser path. Five `#[test]` expected schemas updated accordingly.

### 4. Test tuning: bump `n_messages` in `test_context_shift_with_tool_calls`

The pre-serialized JSON tool strings (point 2) reduced per-render token count enough that 10 messages no longer overflowed `n_ctx=1024` with the standard test model. `context_shift` then skipped, breaking the `assert!(messages_after.len() < messages_before)` check.

Bumped to 20 to restore the overflow condition without changing test intent.

## Stack

- This PR: engine fixes (you are here)
- Next: `feat: Llama-3.x family handler` — independent of this PR
- Next: `feat: LFM2.5 family tool calling (parser-only)` — depends on this PR

### DIFF
--- a/nobodywho/core/src/template.rs
+++ b/nobodywho/core/src/template.rs
@@ -90,7 +90,34 @@ impl ChatTemplate {
             ),
             ("bos_token".to_string(), Value::from(self.bos_token.clone())),
             ("eos_token".to_string(), Value::from(self.eos_token.clone())),
-            ("tools".to_string(), Value::from_serialize(&ctx.tools)),
+            // Pre-serialise each tool to its canonical OpenAI JSON shape (insertion
+            // order preserved by `Tool`'s custom Serialize impl using
+            // `serialize_map`) and pass the list as STRINGS. The HF tool-aware
+            // chat templates (LFM2/LFM2.5/Qwen3 etc.) all guard with
+            // `{%- if tool is not string -%}{%- set tool = tool | tojson -%}{%- endif -%}`
+            // — so a string passes through unchanged. This bypasses minijinja's
+            // internal Value→tojson re-serialisation, which otherwise re-sorts
+            // object keys alphabetically (default `serde_json::Map` is `BTreeMap`).
+            // LFM2.5-350M is empirically sensitive to that order: alphabetical
+            // (`{"function":..., "type":"function"}`) makes it hallucinate kwargs
+            // (`parameters=`, `type=`) in the emitted Pythonic tool call.
+            //
+            // NOTE: `ctx.tools` is `Option<Vec<Tool>>`. `Option::iter` yields 0 or 1
+            // element of type `&Vec<Tool>` — so mapping on it would serialise the
+            // ENTIRE list as a single JSON-array string, producing `[[…]]` after the
+            // template's own `[…]` wrap. Flatten via `as_ref().into_iter().flatten()`
+            // so we map per-Tool.
+            (
+                "tools".to_string(),
+                Value::from_serialize(
+                    ctx.tools
+                        .as_ref()
+                        .into_iter()
+                        .flatten()
+                        .map(|t| serde_json::to_string(t).unwrap_or_default())
+                        .collect::<Vec<_>>(),
+                ),
+            ),
         ];
 
         // Add all template variables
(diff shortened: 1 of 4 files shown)


ahmedibrahim085:
Local verification done before pushing:
- `cargo clippy --no-deps -- -D warnings` clean.
- `cargo build --locked` clean (excluding nobodywho-python which needs maturin's pyo3 setup — CI handles that).
- All 5 `template::tests` pass (including the new `test_render_template_with_tools_preserves_key_order` regression guard for the `tools` template variable wiring).
- pytest `test_tool_calling.py` with Qwen3-0.6B-Q4_K_M: **13/14 pass** (the 1 fail is `test_tool_with_sets` which is model-behavior — Qwen3 swapped `set1`/`set2` arguments — same flake observed on `main` baseline).
- pytest with the same model on the `#521` stack (which adds the terminal Done epilogue): **14/14 pass** — confirms no regression from this PR's engine fixes.

Force-pushed to df766549 (rebased to squash a clippy needless-borrow fix + new test into the original `feat(template)` commit, instead of trailing fixup commits).

nobodywho/core/src/tool_calling/mod.rs
```diff
@@ -209,6 +240,7 @@ pub struct ToolCall {
 }
 
 // Serialize tools according to https://huggingface.co/blog/unified-tool-use
+// (insertion-order preserved — see Serialize for Tool above for rationale).
```
gergelyvagujhelyi: The comment claims insertion order is preserved, but the impl below uses the same serde_json::json!({...}).serialize(serializer) pattern that was just removed from Tool::serialize. Without the serde_json/preserve_order feature (off in core/flutter/uniffi), this serialises alphabetically — verified locally with the default serde_json features:
{"function":{"arguments":{"radius":5},"name":"circle_area"},"type":"function"}
None of the test-fixture templates in template.rs (Llama3 / DeepSeek / Qwen3) do tool_call | tojson on the whole call — they access by attribute, and the only tojson is on tool_call.arguments's model-controlled inner blob. Production chat templates come from the GGUF at runtime via model.chat_template(...) rather than this repo, so strictly this assessment covers the in-repo fixtures + a tree grep, not the LFM2.5 GGUF template itself. Latent under that scope. Two options: apply the same serialize_map / nested helper pattern as Tool::serialize, or rewrite this comment to say "intentionally not preserved — see <reason>".

gergelyvagujhelyi (commented):
Good job! I only found this comment issue.

nobodywho/core/src/template.rs
```diff
@@ -90,7 +90,34 @@ impl ChatTemplate {
             ),
             ("bos_token".to_string(), Value::from(self.bos_token.clone())),
             ("eos_token".to_string(), Value::from(self.eos_token.clone())),
-            ("tools".to_string(), Value::from_serialize(&ctx.tools)),
+            // Pre-serialise each tool to its canonical OpenAI JSON shape (insertion
+            // order preserved by `Tool`'s custom Serialize impl using
+            // `serialize_map`) and pass the list as STRINGS. The HF tool-aware
+            // chat templates (LFM2/LFM2.5/Qwen3 etc.) all guard with
```
gergelyvagujhelyi, line 96: This claim is false for the Qwen3 family. Verified end-to-end against a real production GGUF (Qwen3.6-35B-A3B-UD-Q4_K_M.gguf, arch qwen35moe): its embedded chat template at line 63 of 8057 bytes does unguarded {{ tool | tojson }} — no is not string check, no separate tool_use template. The HF Qwen/Qwen3-8B/tokenizer_config.json and the in-tree fixture at template.rs:433 match.

Rendered through the repo's exact minijinja 2.11.0 (builtins, json, loader) + pycompat with the PR's tools-as-strings wiring, the <tools> block the model receives is:
<tools>
"{\"type\":\"function\",\"function\":{\"name\":\"circle_area\",\"description\":\"...\",\"parameters\":{...}}}"
</tools>
A JSON string literal — the entire tool definition quoted and backslash-escaped. The model sees an opaque blob, not a tool definition.

The previous Vec<Tool> wiring rendered alphabetical-but-valid JSON, which Qwen3 tolerated. So this PR changes Qwen3-family from "tolerated, working" to "double-encoded, broken." test_render_string_qwen3_template doesn't catch it because it passes tools: None (template.rs:442), so the tools branch is never exercised.

Minijinja's built-in tojson alphabetises any Object regardless of Serialize order or Value::from_iter order — verified — so no input-shape change recovers raw-JSON output through an unguarded template. Either a custom tojson filter or load-time guard detection would be needed.

gergelyvagujhelyi: #[test]
    fn test_render_unguarded_template_emits_raw_tool_json_not_string_literal() {
        use std::sync::Arc;

        // Minimal unguarded variant of the template used by
        // `test_render_template_with_tools_preserves_key_order`. Mirrors the
        // shape used by production Qwen3 and the in-tree Qwen3 fixture.
        let template = "{%- for tool in tools -%}{{ tool | tojson }}{%- endfor -%}";

        let tool = Tool::new(
            "circle_area",
            "Compute the area of a circle.",
            serde_json::json!({
                "type": "object",
                "properties": { "radius": { "type": "number" } },
                "required": ["radius"]
            }),
            Arc::new(|_| String::new()),
        );

        let template_vars: HashMap<String, bool> = HashMap::default();
        let ctx = ChatTemplateContext::new(template_vars, Some(vec![tool]));
        let chat_template = ChatTemplate::new(template, "", "").unwrap();
        let rendered = chat_template.render(&[], &ctx).unwrap();

        // Smoking gun: an escaped-quote pair `\"type\":\"function\"`
        // (backslash-quote, not bare quote) in the rendered output means
        // the tool was JSON-encoded as a string literal (double-encoded)
        // rather than emitted as a raw object.
        assert!(
            !rendered.contains(r#"\"type\":\"function\""#),
            "regression: tools rendered as a JSON STRING LITERAL (double-encoded). \
             Unguarded `{{{{ tool | tojson }}}}` over a String value JSON-encodes \
             the value, producing a quoted/escaped form no model can parse as a \
             tool definition. Rendered output:\n{rendered}"
        );

        // The raw OpenAI canonical token must appear — proving the object
        // reaches the prompt as parseable JSON. Key order intentionally not
        // asserted: alphabetical-but-valid JSON was Qwen3's pre-PR working
        // baseline and remains an acceptable fix outcome for unguarded
        // templates.
        assert!(
            rendered.contains(r#""type":"function""#),
            "rendered output is missing the raw `\"type\":\"function\"` token \
             (model would not see a usable tool definition); got:\n{rendered}"
        );
    }    #[test]
    fn test_render_unguarded_template_emits_raw_tool_json_not_string_literal() {
        use std::sync::Arc;

        // Minimal unguarded variant of the template used by
        // `test_render_template_with_tools_preserves_key_order`. Mirrors the
        // shape used by production Qwen3 and the in-tree Qwen3 fixture.
        let template = "{%- for tool in tools -%}{{ tool | tojson }}{%- endfor -%}";

        let tool = Tool::new(
            "circle_area",
            "Compute the area of a circle.",
            serde_json::json!({
                "type": "object",
                "properties": { "radius": { "type": "number" } },
                "required": ["radius"]
            }),
            Arc::new(|_| String::new()),
        );

        let template_vars: HashMap<String, bool> = HashMap::default();
        let ctx = ChatTemplateContext::new(template_vars, Some(vec![tool]));
        let chat_template = ChatTemplate::new(template, "", "").unwrap();
        let rendered = chat_template.render(&[], &ctx).unwrap();

        // Smoking gun: an escaped-quote pair `\"type\":\"function\"`
        // (backslash-quote, not bare quote) in the rendered output means
        // the tool was JSON-encoded as a string literal (double-encoded)
        // rather than emitted as a raw object.
        assert!(
            !rendered.contains(r#"\"type\":\"function\""#),
            "regression: tools rendered as a JSON STRING LITERAL (double-encoded). \
             Unguarded `{{{{ tool | tojson }}}}` over a String value JSON-encodes \
             the value, producing a quoted/escaped form no model can parse as a \
             tool definition. Rendered output:\n{rendered}"
        );

        // The raw OpenAI canonical token must appear — proving the object
        // reaches the prompt as parseable JSON. Key order intentionally not
        // asserted: alphabetical-but-valid JSON was Qwen3's pre-PR working
        // baseline and remains an acceptable fix outcome for unguarded
        // templates.
        assert!(
            rendered.contains(r#""type":"function""#),
            "rendered output is missing the raw `\"type\":\"function\"` token \
             (model would not see a usable tool definition); got:\n{rendered}"
        );
    }

PR #520, opened 2026-05-09
feat(tool_calling): Llama-3.x family handler

ahmedibrahim085:
Splits a chunk of #516 into smaller, independently-reviewable pieces, per @AsbjornOlling's request. This is **PR 2 of 3**.

Refs: #516, #214

## What this PR adds

A grammar-driven `ToolFormat::Llama32` handler covering Llama-3.1, 3.2, and 3.3 models that emit tool calls in the format:

```
<|python_tag|>{"name": "...", "parameters": {...}}<|eot_id|>
```

The `<|python_tag|>` prefix is treated as optional in extraction since the model often omits it.

### Files
- New: `core/src/tool_calling/llama32.rs` — `Llama32Handler` + `generate_grammar` + `extract_tool_calls` + tests.
- Modified: `core/src/tool_calling/mod.rs` — wire `ToolFormat::Llama32` variant + detection-chain entry (template-marker on `Environment: ipython` / `<|python_tag|>`).

### Detection
Template-marker first (covers Llama-3.1+ tool-use templates; Llama-2 templates have neither marker, so no false-positive).

## Independence
This PR does **not** depend on the engine-fixes PR. Llama-3.x handler is grammar-driven, so it's already insulated from the tool-schema-serialization issues that affect smaller parser-only models. Could be reviewed and merged in parallel.

## Stack
- Sibling: engine-fixes PR (independent)
- Sibling: LFM2.5 family tool calling PR (independent of this one)

### DIFF
--- a/nobodywho/core/src/tool_calling/llama32.rs
+++ b/nobodywho/core/src/tool_calling/llama32.rs
@@ -1,0 +1,246 @@
+use super::{Tool, ToolCall, ToolFormatError, ToolFormatHandler};
+use gbnf::builder::{nt, seq, t, GrammarBuilder};
+use gbnf::json::json_schema_to_grammar;
+use gbnf::GbnfGrammar;
+use serde_json::json;
+use tracing::{debug, warn};
+
+#[derive(Debug, Clone, Copy)]
+pub struct Llama32Handler;
+
+const BEGIN: &str = "<|python_tag|>";
+const END: &str = "<|eot_id|>";
+
+#[derive(serde::Deserialize)]
+struct LlamaToolCall {
+    #[serde(alias = "function")]
+    name: String,
+    parameters: serde_json::Value,
+}
+
+impl ToolFormatHandler for Llama32Handler {
+    fn begin_token(&self) -> &str {
+        BEGIN
+    }
+
+    fn end_token(&self) -> &str {
+        END
+    }
+
+    fn generate_grammar(&self, tools: &[Tool]) -> Result<GbnfGrammar, ToolFormatError> {
+        let tool_call_schemas: serde_json::Value = tools
+            .iter()
+            .map(|tool| {
+                json!(
+                    {
+                        "type": "object",
+                        "properties": {
+                            "name": { "const": tool.name },
+                            "parameters": tool.json_schema
+                        },
+                        "required": ["name", "parameters"]
+                    }
+                )
+            })
+            .collect();
+
+        let tool_call_schema = json!({ "oneOf": tool_call_schemas });
+        let json_grammar = json_schema_to_grammar(tool_call_schema, "root")?;
+
+        let grammar = GrammarBuilder::from_existing(json_grammar)
+            .rule(
+                "toolcall",
+                seq(&[t(BEGIN), nt("ws"), nt("root"), nt("ws")]),
+            )
+            .rule("superroot", nt("toolcall"))
+            .root("superroot")
+            .build();
+
+        Ok(grammar)
+    }
+
+    fn extract_tool_calls(&self, input: &str) -> Option<Vec<ToolCall>> {
+        let mut s = input.trim();
+        if let Some(rest) = s.strip_prefix(BEGIN) {
+            s = rest.trim_start();
+        }
+        if let Some(prefix) = s.strip_suffix(END) {
+            s = prefix.trim_end();
+        }
+
+        let mut stream = serde_json::Deserializer::from_str(s).into_iter::<LlamaToolCall>();
+        let first = stream.next()?;
+
+        match first {
+            Ok(call) => {
+                if stream.next().is_some() {
+                    warn!(
+                        "Llama-3.x emitted >1 tool call in one assistant turn; \
+                         only the first is dispatched (chat template constraint)"
+                    );
+                }
+                debug!(tool_name = %call.name, "Successfully parsed Llama tool call");
+                Some(vec![ToolCall {
+                    name: call.name,
+                    arguments: call.parameters,
+                }])
+            }
+            Err(e) => {
+                debug!(error = %e, json = s, "Failed to parse Llama tool call JSON");
+                None
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json::json;
+
+    fn handler() -> Llama32Handler {
+        Llama32Handler
+    }
+
+    #[test]
+    fn p1_round4_with_python_tag_name_field() {
+        let input = r#"<|python_tag|>{"name": "circle_area", "parameters": {"radius": "5"}}"#;
+        let calls = handler().extract_tool_calls(input).unwrap();
+        assert_eq!(calls.len(), 1);
+        assert_eq!(calls[0].name, "circle_area");
+        assert_eq!(calls[0].arguments, json!({"radius": "5"}));
+    }
+
+    #[test]
+    fn p2_round4_no_prefix_name_field() {
+        let input = r#"{"name": "circle_area", "parameters": {"radius": "12"}}"#;
+        let calls = handler().extract_tool_calls(input).unwrap();
+        assert_eq!(calls[0].name, "circle_area");
+        assert_eq!(calls[0].arguments, json!({"radius": "12"}));
+    }
+
+    #[test]
+    fn p3_round4_function_field_alias() {
+        let input = r#"{"function": "get_weather", "parameters": {"city": "Cairo"}}"#;
+        let calls = handler().extract_tool_calls(input).unwrap();
+        assert_eq!(calls[0].name, "get_weather");
+        assert_eq!(calls[0].arguments, json!({"city": "Cairo"}));
+    }
+
+    #[test]
+    fn p4_round4_function_field_alias() {
+        let input = r#"{"function": "get_weather", "parameters": {"city": "London"}}"#;
+        let calls = handler().extract_tool_calls(input).unwrap();
+        assert_eq!(calls[0].name, "get_weather");
+        assert_eq!(calls[0].arguments, json!({"city": "London"}));
+    }
+
+    #[test]
+    fn p5_round4_function_field_alias() {
+        let input = r#"{"function": "get_weather", "parameters": {"city": "Tokyo"}}"#;
+        let calls = handler().extract_tool_calls(input).unwrap();
+        assert_eq!(calls[0].name, "get_weather");
+        assert_eq!(calls[0].arguments, json!({"city": "Tokyo"}));
+    }
+
+    #[test]
+    fn trailing_eot_is_stripped() {
+        let input = r#"<|python_tag|>{"name":"a","parameters":{}}<|eot_id|>"#;
+        let calls = handler().extract_tool_calls(input).unwrap();
+        assert_eq!(calls[0].name, "a");
+    }
+
+    #[test]
+    fn whitespace_between_prefix_and_json_tolerated() {
+        let input = "<|python_tag|>  \n  {\"name\": \"circle_area\", \"parameters\": {\"radius\": \"5\"}}";
+        let calls = handler().extract_tool_calls(input).unwrap();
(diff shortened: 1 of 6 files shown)


nobodywho/core/src/tool_calling/mod.rs
```diff
@@ -379,6 +391,14 @@ pub fn detect_tool_format(model: &LlamaModel) -> Result<ToolFormat, ToolFormatEr
             debug!("Detected Qwen3 format from architecture");
             return Ok(ToolFormat::Qwen3(Qwen3Handler));
         }
+        // Llama-3.x architecture (after Qwen so Qwen-llama derivatives still
```
gergelyvagujhelyi, line 490: The current comment says the marker check would have "rejected" non-3.x llama models, but the marker check returns on success and falls through on no-match. So Llama-2 / Llama-3.0 GGUFs (arch="llama", no  tool-use markers) reach this branch too. Maybe reword to something like:
"Llama-3.x architecture (after Qwen so Qwen-llama derivatives still match Qwen first). Best-effort fallback for llama-arch models that lack tool-use markers. Llama-2 / Llama-3.0 base will land here too and produce degraded output; those configurations aren't expected to work with tools regardless of which handler is selected."

gergelyvagujhelyi: but I'd argue the fix is deletion, not rewording. arch.starts_with("llama") converts a clean Err(UnsupportedFormat) into silent misclassification for FireFunction-v2, DeepSeek-R1-Distill-Llama, Vicuna, and Llama-2 — none of which speak this format. Today the user gets a clear error; with this branch they get garbled output

gergelyvagujhelyi: Verified locally with TinyLlama-1.1B-Chat-v1.0-Q4_K_M (arch=llama, name=tinyllama_…, no Llama-3.x template markers). With the current PR, detect_tool_format returns Ok(Llama32(Llama32Handler)). With the if arch_lower.starts_with("llama") branch deleted, the same input correctly returns Err(UnsupportedFormat) and the rest of the suite (48 passed) is unaffected. The regression test:
#[test]
#[ignore = "requires LLAMA2_MODEL env var pointing at a non-Llama-3.x llama-arch GGUF \
            (Llama-2 / Vicuna / WizardLM / DeepSeek-R1-Distill-Llama / FireFunction-v2 / TinyLlama)"]
fn arch_fallback_misclassifies_non_llama3() {
    // Isolates the `arch.starts_with("llama")` branch: any llama-arch GGUF
    // without <|python_tag|> or `Environment: ipython` template markers and
    // without "3" in general.name reaches that branch and *only* that branch.
    // None of those models speak Llama-3.x tool-call format, so detection
    // should return Err — today it returns Ok(Llama32) and silently breaks.
    let path = std::env::var("LLAMA2_MODEL").expect("set LLAMA2_MODEL");
    let model = crate::llm::get_model(&path, false, None, None).expect("load model");

    // Sanity-check the test setup: confirm we're hitting the arch branch in isolation.
    let tmpl: String = model
        .language_model
        .chat_template(Some("tool_use"))
        .or_else(|_| model.language_model.chat_template(None))
        .ok()
        .and_then(|t| t.to_string().ok())
        .unwrap_or_default();
    assert!(
        !tmpl.contains("<|python_tag|>") && !tmpl.contains("Environment: ipython"),
        "test setup: GGUF has Llama-3.x template markers — pick a Llama-2 / Vicuna / \
         WizardLM / DeepSeek-R1-Distill-Llama / FireFunction-v2 / TinyLlama GGUF instead"
    );

    let result = detect_tool_format(&model.language_model);
    assert!(
        matches!(result, Err(ToolFormatError::UnsupportedFormat(_))),
        "expected Err(UnsupportedFormat) for non-Llama-3.x llama-arch GGUF, got {result:?}. \
         The arch.starts_with(\"llama\") branch is converting a clean error into silent \
         misclassification."
    );
}

nobodywho/core/src/tool_calling/llama32.rs
```diff
@@ -0,0 +1,246 @@
...
+            })
+            .collect();
+
+        let tool_call_schema = json!({ "oneOf": tool_call_schemas });
+        let json_grammar = json_schema_to_grammar(tool_call_schema, "root")?;
+
+        let grammar = GrammarBuilder::from_existing(json_grammar)
+            .rule(
+                "toolcall",
+                seq(&[t(BEGIN), nt("ws"), nt("root"), nt("ws")]),
```
gergelyvagujhelyi: The official Llama-3.x chat template (Llama-3.1-8B-Instruct.jinja:84-89 (https://github.com/ggml-org/llama.cpp/blob/master/models/templates/meta-llama-Llama-3.1-8B-Instruct.jinja#L84-L89)) renders custom tool calls as raw JSON without <|python_tag|> — the prefix only appears in the built-in-tools branch. So the lazy-grammar trigger here rarely fires for typical use, and chat.rs:wrap_respond (which keys its suppressor on the same token) leaks the JSON tokens to streaming consumers (Godot / Flutter Stream<String> / raw Fn(WriteOutput)). Pasted a unit test in chat.rs that reproduces this in isolation — both tests pass on this branch:

gergelyvagujhelyi:
```
// Add inside the existing `#[cfg(test)] mod tests { ... }` block in chat.rs
// (the same block that already has assert_valid_message_structure, etc).
// `wrap_respond` is module-private, so the test must live in chat.rs.

/// Reproduces the streaming JSON-leak path for Llama-3.x with custom tools.
/// Per the official Llama-3.1 / 3.2 / 3.3 chat templates, custom tool calls
/// are rendered as raw JSON without `<|python_tag|>`. `wrap_respond` keys
/// its emit-suppressor on that exact token, so when the model omits the
/// prefix (the common case for custom tools, matching p2-p5 of the PR's
/// bake-off), every JSON token leaks straight to the streaming consumer.
#[test]
fn streaming_leak_when_python_tag_omitted() {
    use std::sync::{Arc, Mutex};

    let received: Arc<Mutex<Vec<String>>> = Arc::new(Mutex::new(Vec::new()));
    let received_clone = Arc::clone(&received);
    let user_callback = move |x: llm::WriteOutput| {
        if let llm::WriteOutput::Token(t) = x {
            received_clone.lock().unwrap().push(t);
        }
    };

    let (mut wrapped, _resp_rx) =
        wrap_respond(user_callback, Some("<|python_tag|>".to_string()));

    let tokens = [
        "{", "\"name\"", ":", "\"get_weather\"", ",",
        "\"parameters\"", ":", "{", "\"city\"", ":", "\"Tokyo\"", "}", "}",
        "<|eot_id|>",
    ];
    for tok in &tokens {
        wrapped(llm::WriteOutput::Token(tok.to_string()));
    }

    let leaked: String = received.lock().unwrap().iter().cloned().collect();
    assert!(
        leaked.contains("\"name\"") && leaked.contains("get_weather"),
        "expected JSON tokens to leak to the streaming consumer because \
         <|python_tag|> never appeared, got: {leaked:?}"
    );
}
```

```
/// Control: when `<|python_tag|>` IS emitted, suppression works as designed.
#[test]
fn streaming_suppressed_when_python_tag_present() {
    use std::sync::{Arc, Mutex};

    let received: Arc<Mutex<Vec<String>>> = Arc::new(Mutex::new(Vec::new()));
    let received_clone = Arc::clone(&received);
    let user_callback = move |x: llm::WriteOutput| {
        if let llm::WriteOutput::Token(t) = x {
            received_clone.lock().unwrap().push(t);
        }
    };

    let (mut wrapped, _resp_rx) =
        wrap_respond(user_callback, Some("<|python_tag|>".to_string()));

    wrapped(llm::WriteOutput::Token("<|python_tag|>".to_string()));
    for tok in ["{", "\"name\"", ":", "\"get_weather\"", "}"] {
        wrapped(llm::WriteOutput::Token(tok.to_string()));
    }

    let leaked: String = received.lock().unwrap().iter().cloned().collect();
    assert!(
        !leaked.contains("\"name\"") && !leaked.contains("get_weather"),
        "expected suppression after <|python_tag|>, got: {leaked:?}"
    );
}
```

ahmedibrahim085:
**Draft pending stack merge.**

Force-pushed `1e2dbf8d` — currently 12 commits = 4 from #519 + 4 from #523 + 4 handler-only. Self-contained for CI + matrix verification on this branch.

Order of merge (recommended):
1. #519 — engine fixes (template + serialization + clippy).
2. #523 — engine parallel-call cap + grammar trigger patterns + Done epilogue.
3. This PR — rebased to **handler-only 4 commits** after the two above land. Will undraft at that point.

Pytest matrix on `1e2dbf8d` (FRESH binary, bypass-uv, fork ahmedibrahim085/nobodywho):
- Qwen3-0.6B-Q4_K_M: 14/14
- Llama-3.2-3B-Instruct-Q4_K_M: **14/14** full pass
- Llama-3.2-1B-Instruct-Q4_K_M: 11/14 (parser-only edge cases on small model)

nobodywho/core/src/tool_calling/llama32.rs
```diff
@@ -0,0 +1,379 @@
...
+            .rule("superroot", nt("toolcall"))
+            .root("superroot")
+            .build();
+
+        Ok(grammar)
+    }
+
+    fn extract_tool_calls(&self, input: &str) -> Option<Vec<ToolCall>> {
+        let mut s = input.trim();
+        if let Some(rest) = s.strip_prefix(BEGIN) {
```
gergelyvagujhelyi, line 93: `strip_prefix` only matches at the very start, so any preamble before `<|python_tag|>` (e.g. `"Sure, let me check.\n<|python_tag|>{...}<|eot_id|>"`) fails to parse — common in Llama-3.x small-model output. One-line fix:

```rust
if let Some(begin_at) = s.find(BEGIN) {
    s = s[begin_at + BEGIN.len()..].trim_start();
}
```

The false-positive guard (`plain_assistant_text_with_inline_json_returns_none`) still holds: absent-marker leaves `s` untouched, so prose like `"Sure, here is some JSON: {...}"` still fails the JSON parse.

Verified locally against the latest revision (1e2dbf8d) — all 20 existing tests pass (the 16 originals + the 4 new `extract_repairs_*` tests), plus this new one:

```rust
#[test]
fn preamble_before_python_tag_is_tolerated() {
    let input = "Sure, let me check.\n<|python_tag|>{\"name\":\"get_weather\",\"parameters\":{\"city\":\"Cairo\"}}<|eot_id|>";
    let calls = handler().extract_tool_calls(input).unwrap();
    assert_eq!(calls[0].name, "get_weather");
    assert_eq!(calls[0].arguments, json!({"city": "Cairo"}));
}
```

nobodywho/core/src/tool_calling/llama32.rs
gergelyvagujhelyi: Update post force-push: the new `grammar_trigger_patterns()` on lines 36-38 closes half of this — the grammar now fires on `{` too, so model output is at least well-formed JSON. But `wrap_respond` (`chat.rs:2102`) still keys its emit-suppressor on `<|python_tag|>` only:

```rust
llm::WriteOutput::Token(tok) if tool_call_begin_token.as_ref() == Some(tok) => {
    emitting = false;
}
```

When the model emits raw JSON without the prefix (the common case per the official Llama-3.x template), the grammar constrains the output but the tokens still flow through to streaming consumers. The two unit tests I posted (`streaming_leak_when_python_tag_omitted` / `streaming_suppressed_when_python_tag_present`) still pass against the updated branch — so the leak is unchanged.

Suggested fix: thread the handler's trigger patterns (or any additional begin marker like `{`) into `wrap_respond` so suppression matches the grammar's coverage.

AsbjornOlling:
Llama3 models are old (>1yr) and easily outperformed by modern models of the same size or less. There's too many things to address in this PR compared the the (pretty small) value of supporting Llama3 models, so I'm gonna close this for now.

If there is some real reason to support llama3 models, instead of just using a better and newer model, please lmk by opening an issue.

PR #521, opened 2026-05-09
feat(tool_calling): LFM2.5 family handler (parser-only)

ahmedibrahim085:
Splits a chunk of #516 into smaller, independently-reviewable pieces, per @AsbjornOlling's request. This is **PR 3 of 3**.

Refs: #516, #214

> **Note:** this PR is stacked on #519 (engine fixes). The 8 commits include #519's 4 commits at the base. Diff against `main` will shrink to just the 4 LFM2.5-specific commits at the tip once #519 lands. Recommend reviewing #519 first.

## What this PR adds

A **parser-only** `ToolFormat::Lfm2_5` handler covering LFM2.5-350M, LFM2.5-1.2B-Instruct, and LFM2.5-1.2B-Thinking.

LFM2.5 emits tool calls as Pythonic call lists between special tokens:

```
<|tool_call_start|>[circle_area(radius=5)]<|tool_call_end|>
```

### Files (LFM2.5-specific commits at the tip — last 4 in the log)
- New: `core/src/tool_calling/lfm2_5.rs` — `Lfm2_5Handler` + Pythonic-literal parser (list/set/dict/tuple/string/int/float/bool) + extraction tests.
- Modified: `core/src/tool_calling/mod.rs` — wire `ToolFormat::Lfm2_5` variant + detection-chain entry on `keep_past_thinking` jinja marker.
- Modified: `core/src/chat.rs` — terminal Done epilogue gated by an emit-tracking flag, with `test_done_emitted_exactly_once_per_ask` regression guard.

### Why parser-only (no GBNF)?

GBNF grammar attached to LFM2.5 makes the sampler stall (500% CPU, no token output for >10 min) because the model's natural distribution doesn't satisfy the grammar accept-set. Documented at https://github.com/ggml-org/llama.cpp/issues/19068. `generate_grammar` returns `Err(ToolFormatError::UnsupportedFormat(...))`. The engine path already tolerates this via `tool_grammar=None / tool_format=Some(handler)` — sampler runs unconstrained, `extract_tool_calls` runs on output.

Mirrors llama.cpp's pre-autoparser working configuration for LFM2.5.

### Detection: `keep_past_thinking` chat-template marker

LFM2.5 templates uniquely contain a `keep_past_thinking` jinja variable. LFM2 v1 templates use `<|tool_list_start|>` wrappers and lack this marker.

We must NOT rely on `general.name` for the 1.2B variants — their GGUFs ship with `general.name` set to a git-commit SHA (e.g. "4cd563d5...", "E7Ae5Ebc..."), bypassing the name-substring fallback. Template-marker detection covers all LFM2.5 GGUFs regardless of the metadata-naming bug.

### Terminal Done epilogue (chat.rs)

Parser-only paths can exit `Worker::ask` after `wrap_respond` suppressed iter-1's Done past the begin_token, with `extract_tool_calls` returning None. Without the terminal epilogue, raw `Fn(WriteOutput)` consumers never receive a Done — surfaces as `WorkerCrashed` (godot binding) or stuck `TokenStream` (Flutter).

The epilogue is gated by a shared `Arc<AtomicBool>` set by `wrap_respond` when it forwards a Done, so grammar paths (Qwen3, Llama-3.x) that already emitted iter-N Done don't double-fire for raw callback consumers. `TokenStream` consumers were already shielded by the `completed_response.is_some()` short-circuit; this fix extends correctness to external callback users.

Regression test `chat::tests::test_done_emitted_exactly_once_per_ask` was contributed by @gergelyvagujhelyi via review on #516 — included with light style adjustments to match `test_chat_worker`.

## On-device validation

LFM2.5-350M, 5-prompt bake-off on a Galaxy S22 (snapshot at `reports/r17-lfm2-5-350m-thinkingoff_evidence/` in the agent workspace):
- p1 NUMERIC_OK (tool fired, `78.54` exact)
- p2 NUMERIC_PRESENT (tool fired, model wrote arithmetic-only summary)
- p3 TOOL_UPSTREAM_ERR (tool fired correctly, open-meteo unreachable from device WiFi — confounder)
- p4/p5 TOOL_PARSE_ERR (model bypassed tool with canned text — model's own choice, not a parse failure)
- 0 stalls, 0 GetterErrorImpl (was 5/5 GetterErrorImpl on r10 baseline)

Strong improvement from r10 (0/5 reach with crash) to r17 (5/5 reach with 2/5 hard OK).

## Stack
- Depends on: #519 (engine fixes) — merge first
- Sibling: #520 (Llama-3.x family handler) — independent

### DIFF
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -2430,9 +2430,14 @@ mod tests {
         test_utils::init_test_tracing();
         let model = test_utils::load_test_model();
 
-        // Use a very small context size to force shifting
+        // Use a very small context size to force shifting. Bumped n_messages
+        // from 10 -> 20 because passing tools to the chat template as
+        // pre-serialized JSON strings (see commit a26dfc79) reduced the
+        // per-render token count enough that 10 exchanges no longer overflow
+        // n_ctx=1024 with the test_chat-model GGUF — context_shift then
+        // skipped, leaving messages_after.len() == messages_before.
         let n_ctx = 1024;
-        let n_messages = 10;
+        let n_messages = 20;
         let mut worker = Worker::new_chat_worker(
             &model,
             ChatConfig {
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -50,7 +50,7 @@ use std::cmp::min;
 use std::collections::HashSet;
 use std::hash::Hasher;
 use std::path::PathBuf;
-use std::sync::atomic::AtomicBool;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, MutexGuard};
 use tracing::{debug, error, info, trace, trace_span};
 
@@ -1643,11 +1643,21 @@ impl Worker<'_, ChatWorker> {
             },
         );
 
+        // Tracks whether any iteration's Done event was forwarded to the outer
+        // callback. Shared across every `wrap_respond` instance built below so
+        // the terminal-Done epilogue can run only when no Done has reached the
+        // consumer yet (e.g. parser-only paths that exit before producing an
+        // iter-2 generation). Keeps raw `Fn(WriteOutput)` consumers — godot
+        // binding, direct ask_channel users — from receiving two Done events
+        // per `ask` call.
+        let done_forwarded = Arc::new(AtomicBool::new(false));
+
         // get the finished response
         let mut response: String = self.wrapped_update_context_and_generate_response(
             sampler.clone(),
             respond.clone(),
             tool_call_begin.clone(),
+            done_forwarded.clone(),
         )?;
 
         // Process tool calls if tool format is configured
@@ -1691,6 +1701,7 @@ impl Worker<'_, ChatWorker> {
                     sampler.clone(),
                     respond.clone(),
                     tool_call_begin.clone(),
+                    done_forwarded.clone(),
                 )?;
             }
         } // Close if let Some(tool_format)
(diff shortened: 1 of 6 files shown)


nobodywho/core/src/chat.rs
```diff
@@ -1691,17 +1701,31 @@ impl Worker<'_, ChatWorker> {
                     sampler.clone(),
                     respond.clone(),
                     tool_call_begin.clone(),
+                    done_forwarded.clone(),
                 )?;
             }
         } // Close if let Some(tool_format)
 
         debug_assert!(tool_call_begin
```
gergelyvagujhelyi, line 1809: This pre-existing assertion is now wrong because of the change below. With LFM2.5 in parser-only mode, the parser-only-failure path reaches this point with response containing begin_token — the assertion's negation. Debug builds would panic before reaching the new done_forwarded re-emit. Please delete the assertion as part of this PR (the new control flow already makes it a contradiction).

gergelyvagujhelyi: let tool_call_begin = Some("<|tool_call_start|>".to_string());                                                                                                                                                                           
let response = "<|tool_call_start|>[broken_pythonic]<|tool_call_end|>";
 
// chat.rs:1709-1711 verbatim:                                                                                                                                                                                                           
let assertion_holds = tool_call_begin                                                                                                                                                                                                    
     .as_ref()                                                                                                                                                                                                                            
     .is_none_or(|t| !response.contains(t.as_str()));    

assert_eq!(assertion_holds, false); // ← debug build panics here                                                                                                                                                                         

The new if !done_forwarded.load(...) re-emit at Line 1725 is specifically designed to fire on this exact input shape. The assertion above forbids it. Resolution: remove the debug_assert!.

gergelyvagujhelyi:
Also thanks for fixing my original double-Done finding — the flag-based fix is a clean, and the regression test came through neatly with your polish.

ahmedibrahim085:
**Draft pending stack merge.**

Force-pushed `d566379b` — currently 12 commits = 4 from #519 + 4 from #523 + 4 handler-only. Self-contained for CI + matrix verification on this branch.

Order of merge (recommended):
1. #519 — engine fixes (template + serialization + clippy).
2. #523 — engine parallel-call cap + grammar trigger patterns + Done epilogue.
3. This PR — rebased to **handler-only 4 commits** after the two above land. Will undraft at that point.

Pytest matrix on `d566379b` (FRESH binary, bypass-uv, fork ahmedibrahim085/nobodywho):
- Qwen3-0.6B-Q4_K_M: 13/14 (1 RNG flake on `test_tool_with_sets`)
- LFM2.5-1.2B-Instruct-Q4_K_M: **14/14** full pass
- LFM2.5-350M-Q4_K_M: 9/14 (model-size limit on complex prompts)

AsbjornOlling:
Superseded by #564

PR #522, opened 2026-05-09
fix(chat): context_shift gracefully handles single-user-message history

ahmedibrahim085:
## Summary

`Worker::context_shift` (chat.rs) hard-fails with `ShiftError::Message("No deletable messages")` when the chat history contains exactly one user message but the rendered context exceeds `n_ctx`. This shape is the empirical failure mode of small models looping on a tool-call dispatch until `n_ctx` is exhausted with one user prompt + many assistant/tool turns.

This PR adds a fallback path: when no second user message exists, drain the older assistant/tool turns between the first user message and the most recent message, leaving the in-progress tail intact. The standard rolling-pair path (≥ 2 user messages) is unchanged.

## Reproduction (pre-fix)

`pytest tests/test_tool_calling.py::test_tool_with_sets` against `Llama-3.2-1B-Instruct-Q4_K_M.gguf`:

```
ERROR nobodywho.chat:chat.rs:318 Worker crashed:
  Error reading string: Error creating response:
  Error while generating response:
  Error during context shift: Missing expected message No deletable messages
```

Trace: model dispatches `set_intersection` → tool returns → model dispatches AGAIN with stringified args → tool returns → context fills to 1044 tokens (`n_ctx=1024`) → `context_shift` triggered → only 1 user message in history → hard fails.

The predicate (`chat.rs:1383-1385`) and error message (`chat.rs:1388-1391`) predate this PR; git blame: `s194042 2025-10-06`, `Marek Hradil jr 2025-11-07`. The fall-back path in this PR preserves the existing rolling-pair semantics for the multi-user case unchanged.

## What changed

- `nobodywho/core/src/chat.rs::Worker::context_shift` — when `find_next_user_message(first_user_message_index + 1)` returns `None`, use `(first_user_message_index + 1, messages.len() - 2)` as deletable bounds and degrade `find_next_user_message` to fall through to `target_delete_index` instead of `?`-propagating an error.
- `chat::tests::test_context_shift_with_only_one_user_message` regression guard — pre-fills 1 user msg + 30 assistant turns, asserts `context_shift` returns `Ok` and history shrinks (or doesn't grow) while preserving the system prompt and the first user message.

## Test plan

- [x] \`cargo test --release --lib -p nobodywho chat::tests\` — 14 passed (4 context_shift tests including new regression)
- [x] \`cargo clippy --no-deps -p nobodywho -- -D warnings\` — clean
- [x] Manual: \`TEST_MODEL=Qwen3-0.6B pytest test_tool_calling.py\` — 14/14 (no regression)

## Note on scope

This addresses ONLY the "No deletable messages" predicate path. A separate failure mode `NoKvCacheSlot` (chat.rs:1492 guard uses \`==\` instead of \`>=\`, plus missing token-budget recheck after \`read_chunks\`) surfaces on Llama-3.2-3B during longer tool-call loops; it is independent and will be addressed in a follow-up PR.

### DIFF
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -1380,15 +1380,37 @@ impl Worker<'_, ChatWorker> {
                 .ok_or(ShiftError::Message(
                     "No first user message in chat history".into(),
                 ))?;
-        let first_deletable_index = self
+
+        // The standard rolling-pair strategy preserves the last 2 user messages
+        // and shifts older user/assistant pairs out. When the chat has only one
+        // user message — for example a tool-call loop where the model dispatched
+        // the same tool many times without further user input, exhausting n_ctx —
+        // there is no second user message and the previous implementation hard
+        // -failed with "No deletable messages". Fall back to a simpler scheme:
+        // drain the older assistant/tool turns between the first user message
+        // and the most recent message, preserving the in-progress tail.
+        let (first_deletable_index, mut last_deletable_index) = match self
             .find_next_user_message(&messages, first_user_message_index + 1)
-            .ok_or(ShiftError::Message("No deletable messages".into()))?; // Assuming assistant after user
-        let mut last_deletable_index = self
-            .find_start_of_last_n_user_messages(&messages, 2)
-            .ok_or(ShiftError::Message(
-                "Less than two user messages in chat history.".into(),
-            ))?
-            - 1;
+        {
+            Some(second_user_idx) => (
+                second_user_idx,
+                self.find_start_of_last_n_user_messages(&messages, 2)
+                    .ok_or(ShiftError::Message(
+                        "Less than two user messages in chat history.".into(),
+                    ))?
+                    - 1,
+            ),
+            None => {
+                // Need at least: [system?, user, ...middle..., tail] for there to
+                // be anything safe to delete. If the history is too short, leave
+                // it intact and let the upstream caller decide what to do.
+                if messages.len() < first_user_message_index + 3 {
+                    self.extra.messages = messages;
+                    return Ok(());
+                }
+                (first_user_message_index + 1, messages.len() - 2)
+            }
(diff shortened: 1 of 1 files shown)


nobodywho/core/src/chat.rs
```diff
@@ -2825,4 +2848,55 @@ mod tests {
...
+        }
+
+        let messages_before = worker.extra.messages.len();
+        worker.context_shift()?; // must not return Err here
+        let messages_after = worker.extra.messages.len();
+
+        assert!(
+            messages_after <= messages_before,
+            "context_shift must not grow the history (before={messages_before}, after={messages_after})"
+        );
```
gergelyvagujhelyi, line 2892: run cargo fmt --all, CI will fail otherwise.

nobodywho/core/src/chat.rs
```diff
@@ -1380,15 +1380,37 @@ impl Worker<'_, ChatWorker> {
...
+            ),
+            None => {
+                // Need at least: [system?, user, ...middle..., tail] for there to
+                // be anything safe to delete. If the history is too short, leave
+                // it intact and let the upstream caller decide what to do.
+                if messages.len() < first_user_message_index + 3 {
+                    self.extra.messages = messages;
+                    return Ok(());
+                }
+                (first_user_message_index + 1, messages.len() - 2)
```
gergelyvagujhelyi, line 1411: Fallback can leave [Sys, User, Tool]
I am not sure if we need to solve this, probably not...

AsbjornOlling:
This PR highlights a real failure mode, but I feel that the suggested solution adds too much complication to the tool-calling logic. It becomes really difficult to reason about. I would like to figure out how to do this in a much more general way.

On a side-note: the description for this PR is heavily vibe-coded and extremely difficult to reason about. For future PRs, please try to understand the problem yourself and explain the situation in plain terms.

PR #523, opened 2026-05-09
feat(engine): parallel tool-call cap + grammar trigger patterns + Done epilogue

ahmedibrahim085:
## Summary

Engine-side infrastructure factored out of #520 / #521 for focused review. Four atomic commits, independent of #519 for compilation.

## Commits

| SHA | Title |
|---|---|
| `da9101bf` | feat(chat): terminal Done epilogue for parser-only paths |
| `f25a1d36` | feat(engine): pattern-based lazy grammar trigger via ShiftStep::GrammarPattern |
| `732e2fa7` | feat(engine): per-handler allow_repeated_calls policy |
| `b7a8f8ac` | fix(chat): context_shift gracefully handles single-user-message history |

## Why factor out

#520 (Llama-3.x handler) and #521 (LFM2.5 handler) both override `grammar_trigger_patterns()` and `allow_repeated_calls()` trait methods + need the terminal Done epilogue for parser-only paths. Keeping engine infra in handler PRs forced redundant review surface and cross-PR conflicts.

## Verification

- `cargo build --release` clean.
- `cargo test --lib tool_calling` 31 pass, 1 ignored.
- `cargo test --lib chat::tests` 16 pass.
- `b7a8f8ac` is byte-identical to #522 commit `76613a9f` (cherry-picked).

## Stack

This PR is independent of #519 for compilation. Both should land before #520 / #521 rebase to handler-only diffs.

### DIFF
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -50,7 +50,7 @@ use std::cmp::min;
 use std::collections::HashSet;
 use std::hash::Hasher;
 use std::path::PathBuf;
-use std::sync::atomic::AtomicBool;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, MutexGuard};
 use tracing::{debug, error, info, trace, trace_span};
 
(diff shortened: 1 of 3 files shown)


gergelyvagujhelyi:
This seems functionally correct, I would add more meaningful test personally but I might do it later, since you have a lot of PR I would appreciate your effort and let this one in, so I approved the pipeline but the CI fails pls run `cargo fmt` to apply code auto-formatting

nobodywho/core/src/chat.rs
```diff
@@ -1380,15 +1380,37 @@ impl Worker<'_, ChatWorker> {
...
+                    ))?
+                    - 1,
+            ),
+            None => {
+                // Need at least: [system?, user, ...middle..., tail] for there to
+                // be anything safe to delete. If the history is too short, leave
+                // it intact and let the upstream caller decide what to do.
+                if messages.len() < first_user_message_index + 3 {
+                    self.extra.messages = messages;
+                    return Ok(());
```
gergelyvagujhelyi, line 1409: @AsbjornOlling is this a type of silent failure we should fix? Maybe return this: Err(ShiftError::Message(format!(... ?

nobodywho/core/src/chat.rs
```diff
@@ -2294,6 +2475,129 @@ mod tests {
...
+            Ok(_) => {
+                let resp = receiver.recv_timeout(std::time::Duration::from_secs(120));
+                println!("=== ask returned Ok, response ===");
+                match resp {
+                    Ok(s) => println!("{s}"),
+                    Err(e) => println!("recv error (worker dropped without Done?): {e}"),
+                }
+            }
+            Err(e) => {
+                println!("=== ask returned Err ===");
```
gergelyvagujhelyi, line 2587: The body uses `println!` rather than asserting — so even when run manually with `--ignored`, a future regression would just print `ask returned Err <WorkerCrashed>` and the test still passes.
But since it is inside an ignored It doesn't matter that much...

PR #525, opened 2026-05-11
Nobodywho Flutter 2.1.0 and a bunch of other releases

AsbjornOlling:


gergelyvagujhelyi:
Random question: Per-binding GitHub Releases for python/godot/RN already exist but have empty bodies — should we fill those in with the relevant changes too?

PR #526, opened 2026-05-13
Wasm pr1 foundation

gergelyvagujhelyi:
## Description

Lays the foundation for a new WebAssembly binding for NobodyWho, letting downstream consumers run local LLMs in a browser tab via the same core engine that powers the Python, Godot, Flutter, and Uniffi bindings. This PR is the first of a stack — it gets the workspace compiling for `wasm32-unknown-emscripten` and exposes a minimal JS surface; richer features (CrossEncoder, Constraint/structured output, browser demo, release pipeline) come in follow-up PRs.

Three threads of work:

1. **New `nobodywho/wasm/` crate** — async-only JS surface mirroring the Python binding. Exposes `Model` (loaded from in-memory `Uint8Array`), `Chat`, `TokenStream`, `Encoder` via `wasm-bindgen`. Uses `crate-type = ["cdylib", "rlib"]` so `cargo check --workspace` keeps working without the wasm toolchain installed. Each exported method returns `js_sys::Promise` (rather than `pub async fn`) to sidestep the `UnwindSafe` bound that `wasm_bindgen_futures::future_to_promise` otherwise imposes; a `promisify()` helper wraps bodies in `AssertUnwindSafe + catch_unwind` so a panic surfaces as a rejected Promise instead of tearing down the wasm instance.

2. **`core` cfg-gating for wasm32** — splits dependencies and code paths so the crate compiles on `wasm32-unknown-emscripten`:
   - `tokio` split: `rt-multi-thread` on native, plain `rt` on wasm.
   - `ureq` (raw sockets), `indicatif` (terminal UI), `monty` (Python interpreter) and `bashkit` (virtual bash) gated to non-wasm.
   - `dirs` gated to non-android-non-wasm.
   - `jsonschema` default features disabled in `grammar/gbnf` (drops `reqwest`/`hyper`/`mio` chain — the only call site is `jsonschema::meta::is_valid` which doesn't need HTTP/file resolution).
   - The entire model-loader/HTTP-cache/path-parser pipeline (`get_model`, `get_model_async`, `download_*`, `parse_model_path`, `get_cache_dir`) is gated to non-wasm; `Tool::python` and `Tool::bash` likewise.
   - The synchronous `chat::TokenStream` and four load-planning items in `core/src/memory.rs` (`GgufModelInfo`, `read_gguf_model_info`, `estimate_per_layer_bytes`, `plan_model_loading`) are also gated — they're reachable only from the gated native loader, and gating them avoids wasm-target dead-code warnings. `plan_context` / `ContextPlan` stay public on all targets since chat-context setup uses them on wasm too.

3. **Worker refactor** — channel switched from `std::sync::mpsc` to `tokio::sync::mpsc::unbounded_channel` so the worker can run as either a real OS thread (native, via `blocking_recv()`) or a `wasm_bindgen_futures::spawn_local` future on the JS event loop (wasm, via `recv().await`). `WorkerGuard` grew a wasm-only constructor that omits the `JoinHandle` and a cfg-gated Drop tail. Affects `ChatHandleAsync`, `EncoderAsync`, `CrossEncoderAsync`. The synchronous `ChatHandle` is gated to non-wasm (no blocking in a browser tab).

Also adds `pub fn get_model_from_bytes(bytes: &[u8], gpu_layers: u32) -> Result<Model, LoadModelError>` in `core/src/llm.rs` — a path-free model loader required by wasm (no filesystem) and useful on native for tests where the GGUF is already in memory. Gated off Windows MSVC because the underlying `LlamaModel::load_from_buffer` uses POSIX `fmemopen`. Covered by a native smoke test that loads `TEST_MODEL` into a `Vec<u8>` and asserts a successful round-trip.

The `llama-cpp-2` git source switched from `marek-hradil/llama-cpp-rs#main` to `nobodywho-ooo/llama-cpp-rs#wasm`. That fork carries Marek's existing llguidance/EOS-fix patches plus the Emscripten `TargetOs` variant, sysroot auto-detection, bindgen configuration, and the cmake gating that disables every GPU backend on wasm builds.

`Cargo.nix` and `crate-hashes.json` are regenerated against this PR's `Cargo.lock` so `nix flake check` succeeds. **Two nix-side gotchas worth noting for follow-up PR authors in this stack** (each pr2/3/4 will need its own end-of-PR regen as well, because each touches Cargo.toml/Cargo.lock):

- `crate2nix`'s `-h crate-hashes.json` flag caches hashes by `name@version`, not by full URL+commit. A URL change (like marek-hradil → nobodywho-ooo) without manually updating the cache entry silently reuses the old tree's hash, and `nix flake check` then fails on a fetchgit hash mismatch.
- `pkgs.fetchgit` defaults `fetchSubmodules = true`, while `nix-prefetch-git` defaults to `--no-fetch-submodules`. The `llama-cpp-rs` repo vendors `llama.cpp` as a submodule, so prefetching without the matching `--fetch-submodules` flag produces a hash that doesn't match what fetchgit actually downloads at build time.

Fixes # (issue)

## Type of change

- [ ] Documentation update

Native consumers (Python, Godot, Flutter) see no API changes. The internal `WorkerGuard` channel type changed from `std::sync::mpsc::Sender` to `tokio::sync::mpsc::UnboundedSender`, but `WorkerGuard::send`'s public signature is unchanged.

## How Has This Been Tested?

Local checks against the PR tip:

- `cargo check -p nobodywho` on native (macOS aarch64): clean.
- `cargo check --target wasm32-unknown-emscripten -p nobodywho-wasm`: builds end-to-end through bindgen + cc + cmake. **0 warnings.**
- `cargo fmt --all --check`: clean.
- `cargo clippy --no-deps -- -D warnings` from `nobodywho/core/`: clean (the same invocation the `linting / lint` CI job runs).

Native smoke test for the new `get_model_from_bytes` API, runnable with a model:

```bash
cd nobodywho
export TEST_MODEL=/path/to/model.gguf
cargo test -p nobodywho get_model_from_bytes_loads_a_real_gguf -- --nocapture
```

Not verified end-to-end: actual wasm linking with `emcc`, loading a real GGUF via `Model.loadBytes` in a browser. Browser-side end-to-end verification lands in the demo PR later in the stack.

To reproduce the local checks:

```bash
# Native check (no special setup):
cd nobodywho
cargo check -p nobodywho

# Wasm check (requires emsdk activated for full build, but cargo check works without it):
rustup target add wasm32-unknown-emscripten
cargo check --target wasm32-unknown-emscripten -p nobodywho-wasm
```

For a real wasm build, see `nobodywho/wasm/README.md` for the emsdk install steps.

## Checklist:

### DIFF
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -280,8 +280,8 @@ impl ChatBuilder {
 
     /// Build a blocking chat handle and start the background worker.
     ///
-    /// Native-only. wasm32 has no real threads and no way to `.completed()`
-    /// the result; use `build_async` and await on the returned future instead.
+    /// Native-only. Uses `std::thread::spawn` internally, which doesn't
+    /// work on wasm without thread support enabled. Use `build_async` on wasm.
     #[cfg(not(target_arch = "wasm32"))]
     pub fn build(self) -> ChatHandle {
         ChatHandle::new(self.model, self.config)
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -878,11 +878,17 @@ impl ChatHandleAsync {
 }
 
 /// A stream of tokens from the model.
+///
+/// Native-only synchronous stream — `next_token` calls `blocking_recv`, which
+/// would deadlock the single-threaded JS event loop. Wasm consumers use
+/// [`TokenStreamAsync`] instead, returned from [`ChatHandleAsync::ask`].
+#[cfg(not(target_arch = "wasm32"))]
 pub struct TokenStream {
     rx: tokio::sync::mpsc::UnboundedReceiver<llm::WriteOutput>,
     completed_response: Option<String>,
 }
 
+#[cfg(not(target_arch = "wasm32"))]
 impl TokenStream {
     fn new(rx: tokio::sync::mpsc::UnboundedReceiver<llm::WriteOutput>) -> Self {
         Self {
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -279,6 +279,10 @@ impl ChatBuilder {
     }
 
     /// Build a blocking chat handle and start the background worker.
+    ///
+    /// Native-only. wasm32 has no real threads and no way to `.completed()`
+    /// the result; use `build_async` and await on the returned future instead.
+    #[cfg(not(target_arch = "wasm32"))]
     pub fn build(self) -> ChatHandle {
         ChatHandle::new(self.model, self.config)
     }
@@ -292,10 +296,13 @@ impl ChatBuilder {
 /// Interact with a ChatWorker in a blocking manner.
 ///
 /// Use [`ChatBuilder`] to create a new instance with a fluent API.
+/// Native-only synchronous chat handle. Use `ChatHandleAsync` on wasm.
+#[cfg(not(target_arch = "wasm32"))]
 pub struct ChatHandle {
     guard: WorkerGuard<ChatMsg>,
 }
 
+#[cfg(not(target_arch = "wasm32"))]
 impl ChatHandle {
     /// Create a new chat handle directly. Consider using [`ChatBuilder`] for a more ergonomic API.
     pub fn new(model: Arc<llm::Model>, config: ChatConfig) -> Self {
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -299,7 +299,7 @@ pub struct ChatHandle {
 impl ChatHandle {
     /// Create a new chat handle directly. Consider using [`ChatBuilder`] for a more ergonomic API.
     pub fn new(model: Arc<llm::Model>, config: ChatConfig) -> Self {
-        let (msg_tx, msg_rx) = std::sync::mpsc::channel();
+        let (msg_tx, mut msg_rx) = tokio::sync::mpsc::unbounded_channel();
 
         let should_stop = Arc::new(AtomicBool::new(false));
         let should_stop_clone = Arc::clone(&should_stop);
@@ -313,7 +313,10 @@ impl ChatHandle {
                 }
             };
 
-            while let Ok(msg) = msg_rx.recv() {
+            // `blocking_recv` on a tokio unbounded channel is safe outside a
+            // tokio runtime context (we're in a raw `std::thread::spawn`'d
+            // thread). Returns `None` when the channel is closed.
+            while let Some(msg) = msg_rx.blocking_recv() {
                 if let Err(e) = process_worker_msg(&mut worker_state, msg) {
                     return error!("Worker crashed: {e}");
                 }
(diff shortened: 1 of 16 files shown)


AsbjornOlling:
I wonder how using `wasm_bindgen_futures::spawn_local` for the worker will affect the main JS thread.
I fear that doing compute-bound stuff on the main thread will block other things from running, making UI unresponsive, but idk exactly what the properties are.

nobodywho/wasm/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+            handle: builder.build_async(),
+        })
+    }
+
+    /// Send a prompt and receive a `TokenStream`. Tokens arrive as they're
+    /// generated; await `nextToken()` in a loop, or call `completed()` to
+    /// resolve to the full response.
+    pub fn ask(&self, prompt: String) -> js_sys::Promise {
+        let handle = self.handle.clone();
+        promisify(async move {
```
AsbjornOlling, line 143: Does this need to be async? Seems like nothing is awaited inside the async closure.

nobodywho/wasm/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+            let stream = handle.ask(prompt);
+            Ok(TokenStream {
+                inner: Arc::new(tokio::sync::Mutex::new(stream)),
+            })
+        })
+    }
+
+    /// Reset the conversation. Optionally provide a new system prompt.
+    /// Tools are cleared on reset.
+    pub fn reset(&self, system_prompt: Option<String>) -> js_sys::Promise {
```
AsbjornOlling, line 153: Why not pass in tools here?

nobodywho/wasm/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "nobodywho-wasm"
```
AsbjornOlling, line 2: Maybe we could come up with a better name for this crate?
If we will eventually support wasm target for more bindings (e.g. flutter), then naming the vanilla Js bindings crate "wasm" might be a bit misleading.

I'll suggest naming this crate after the language bindings that we're targetting. E.g. "javascript" or "typescript" or "js-web" or something like that.

nobodywho/core/src/chat.rs
```diff
@@ -279,6 +279,10 @@ impl ChatBuilder {
     }
 
     /// Build a blocking chat handle and start the background worker.
+    ///
+    /// Native-only. wasm32 has no real threads and no way to `.completed()`
```
AsbjornOlling: Hmm.. is this true? It definitely is possible to `.completed()`, and web workers should work like threads, no?

nobodywho/core/src/chat.rs
```diff
@@ -279,6 +279,10 @@ impl ChatBuilder {
...
+    /// the result; use `build_async` and await on the returned future instead.
+    #[cfg(not(target_arch = "wasm32"))]
```
AsbjornOlling, line 285: Do we need to feature-gate everything that's not used in the current bindings?

gergelyvagujhelyi:
> I wonder how using `wasm_bindgen_futures::spawn_local` for the worker will affect the main JS thread. I fear that doing compute-bound stuff on the main thread will block other things from running, making UI unresponsive, but idk exactly what the properties are.

This is the trade-off pr1 makes. The follow-up in pr3 adds two things on top of this foundation:

    A per-token streaming hook in core/src/llm.rs — the wasm binding fires a JS callback inside the decode loop.
    A browser-chat-worker.html example that runs nobodywho inside a Web Worker and uses the streaming hook to postMessage(token) back to main as tokens arrive.

nobodywho/wasm/src/lib.rs
gergelyvagujhelyi: No but every Chat method returns a Promise from JS, so await chat.X(...) works the same way everywhere.

gergelyvagujhelyi: I can drop it if you'd rather have it sync.

nobodywho/wasm/src/lib.rs
gergelyvagujhelyi: Tools support comes in a later PR.

nobodywho/wasm/Cargo.toml
gergelyvagujhelyi: nobodywho-js

gergelyvagujhelyi: ?

nobodywho/core/src/chat.rs
gergelyvagujhelyi: std::thread::spawn doesn't work on wasm without -pthread + SharedArrayBuffer, which this binding doesn't enable. I'll refine the doc.

nobodywho/core/src/chat.rs
gergelyvagujhelyi: we can take a look at it

PR #531, opened 2026-05-13
feat: WebAssembly binding (full stack)

gergelyvagujhelyi:
## WebAssembly binding (full stack)

Runs NobodyWho's full core engine — chat + per-token streaming, tool calling, structured output, embeddings, reranking, vision — in the browser (or any wasm host), via llama.cpp + Emscripten, exposed to JS through `wasm-bindgen`. Same engine as the Python/Godot/Flutter/Uniffi bindings.

- Multi-threaded inference (Emscripten pthreads); `modelUrl` (browser) / `modelPath` (Node).
- **wasm32** (4 GiB) and **wasm64 / MEMORY64** (16 GiB, for models whose working set exceeds 4 GiB).
- Built on two `nobodywho-ooo` forks (`wasm-bindgen`, `emscripten`) for Emscripten output mode + pthread / SharedArrayBuffer / NODEFS fixes.

Consolidates #526 / #529 / #530 (all commits preserved). API + setup details in `nobodywho/js/README.md`.


nobodywho/js/examples/browser.html
```diff
@@ -0,0 +1,130 @@
...
+
+  <p>
+    <label>GGUF file: <input id="model" type="file" accept=".gguf" /></label><br />
+    <label>Text: <input id="text" value="Hello from the browser." size="40" /></label><br />
+    <button id="go">Run</button>
+  </p>
+
+  <pre id="out"></pre>
+
+  <script type="module">
```
AsbjornOlling: Seems like there is quite a lot of boilerplate code in this example. Is the idea that downstream users should write all of this to run NW in the browser? I think that's too complicated of an API.

nobodywho/core/src/tokenizer.rs
```diff
@@ -18,14 +18,33 @@ use tracing::{info, warn};
...
+/// Get the media-marker string used to split text from media in rendered
+/// chat templates. Native reads this from llama.cpp's
+/// `mtmd_default_marker()`. The cfg gate routes all wasm32 targets to the
+/// inlined literal (`"<__media__>"`) — necessary for wasm32-unknown-unknown
+/// (which can't resolve the mtmd C++ symbol) and harmless for the other
+/// wasm32 OSes like Emscripten, since the literal is the same string the
+/// FFI call would return.
+#[inline]
+fn mtmd_marker_string() -> String {
+    #[cfg(target_arch = "wasm32")]
```
AsbjornOlling: It looks like the wasm builds opt out of the mtmd feature. Does this do anything?

nobodywho/core/src/llm.rs
```diff
@@ -11,15 +12,71 @@ use llama_cpp_2::llama_batch::LlamaBatch;
...
+#[cfg(not(target_arch = "wasm32"))]
 use std::time::Duration;
-use tracing::{debug, debug_span, error, info, info_span, warn};
+use tracing::{debug, debug_span, error, info, warn};
+// `info_span` is only used inside the native model-loader path; gate it so the
+// wasm build doesn't warn about an unused import.
+#[cfg(not(target_arch = "wasm32"))]
+use tracing::info_span;
+
+// --- Streaming hook (wasm32 only) ----------------------------------------
```
AsbjornOlling: This stuff seems terribly complex. Is it really necessary?
tbh it seems a bit whack to have two parallel mechanisms for emitting tokens from the worker.
Why exactly is this necessary? Is it becuase the worker is running as an async promise and would otherwise block the main thread? This is all a bit convoluted.

nobodywho/core/src/llm.rs
```diff
@@ -153,13 +215,71 @@ pub fn has_gpu_backend() -> bool {
...
+/// mmproj path also needs a buffer-based loader; tracked in WASM.md.
+///
+/// # Platform support
+///
+/// Available everywhere except Windows MSVC, mirroring
+/// [`llama_cpp_2::model::LlamaModel::load_from_buffer`]. The underlying
+/// `fmemopen` is POSIX and not in the MSVC CRT; a Windows tempfile
+/// fallback can be added later if a caller needs it.
+#[cfg(not(all(target_os = "windows", target_env = "msvc")))]
+pub fn get_model_from_bytes(bytes: &[u8], gpu_layers: u32) -> Result<Model, LoadModelError> {
```
AsbjornOlling: Idk if this is the right API to expose for using nobodywho on web.
I mainly wrote the `load_from_buffer` handler in my llama-cpp-rs PR just to quickly test it.
I'm not sure the `LlamaModel::load_from_buffer` needs to end up in the final upstream PR.

We probably expect the webapp to download the model on first load. Might it be cleaner to let NW handle that, and stream into the virtual fs, instead of expecting devs to get it themselves with `fetch()` and have NW copy it onto the virtual fs after? As far as I remember, `load_from_buffer` just writes the file into virtual fs.

This might also cause double the memory usage, until the fetch'ed buffer is GC'ed?

nobodywho/js/src/lib.rs
```diff
@@ -0,0 +1,500 @@
...
+/// ```
+// `deny_unknown_fields` matches ConstraintSpec below and surfaces JS-side
+// typos / unsupported options (e.g. `tools: [...]`) as a JsError at
+// construction time, rather than serde silently dropping the field.
+#[derive(serde::Deserialize, Default)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+struct ChatOptions {
+    context_size: Option<u32>,
+    system_prompt: Option<String>,
+    constraint: Option<ConstraintSpec>,
```
AsbjornOlling: This way of specifying constraints doesn't match existing bindings. Let's do it on the sampler level.

PR #532, opened 2026-05-15
feat: Improve error messages - model loading, model downloading and context shifting

marek-hradil:
I have brought in `miette` to handle errors nicely, not giving the full Rust stack trace every time.
This is applied now to the user errors, regarding model loading and context shifting.
Summary of the errors implemented:
 - Model not found: now distinguishes between a missing directory vs. a missing file within an existing directory, with the resolved path shown in each case
- Not a GGUF file: validates the magic bytes before loading; gives a clear error with a link to find GGUF models
- Is a directory: explicit error when a directory path is passed instead of a .gguf file
- Permission denied: surfaces the underlying OS permission error instead of a generic load failure
- Insufficient memory: detects llama.cpp's null-result OOM and reports it as InsufficientMemoryForModel with a suggestion to try a smaller quantization
- Download HTTP errors: HTTP 401, 403, and 404 are now distinct errors with actionable help text (e.g. how to pass a HF token, how to accept a model license)
- Encoder passed to Chat: detects embedding/encoder models (via pooling type metadata) and returns NotAnLLM with a hint to use Encoder instead
- No chat template: replaces a wall-of-text error with a short message and a help: hint
- Input exceeds context: explicit InputExceedsContext error when the tokenized prompt is larger than n_ctx, instead of a debug assertion panic
- Context shift failures: ShiftError::Message(String) split into NoUserMessages and TooFewMessages with actionable help text
- Worker errors propagated: inference errors (e.g. context shift during generation) are now sent back through the TokenStream channel and surfaced as CompletionError::WorkerError instead of being silently dropped; build() / build_async() now return Result so init failures are no longer swallowed


nobodywho/core/src/chat.rs
```diff
@@ -868,20 +884,25 @@ impl TokenStream {
                     self.completed_response = Some(completed_response);
                     return None;
                 }
+                llm::WriteOutput::Error(e) => {
+                    self.error = Some(e);
```
AsbjornOlling: Seems like errors picked up by `TokenStream` are just swallowed in some cases. `completed()` will return the error, but  `.next_token()` returns None. Should we somehow surface this error earlier? 

Maybe we could raise it from `.next_token()` and propagate it up in `.completed()`?

...or maybe I'm just not grokking the error flow.

nobodywho/core/src/llm.rs
```diff
@@ -114,6 +114,28 @@ pub struct Model {
...
 }
 
+impl Model {
+    /// Returns true if this model is an autoregressive LLM capable of text generation.
+    ///
+    /// LLMs never pool token representations, so `<arch>.pooling_type` is absent from
+    /// their GGUF metadata (giving `Unspecified`). Encoder models (BERT, nomic-bert, etc.)
+    /// always have this key set to CLS, Mean, or similar — a reliable, architecture-agnostic
+    /// signal that the model cannot generate text.
+    pub fn is_llm(&self) -> bool {
```
AsbjornOlling: Maybe this is a bit of a misnomer? Surely embeddings models are LLMs too. Maybe "is generative model"?

nobodywho/core/src/llm.rs
```diff
@@ -263,6 +286,11 @@ pub fn get_model(
     let language_model =
         LlamaModel::load_from_file(&LLAMA_BACKEND, &real_model_path, &model_params).map_err(
             |e| {
+                if e.to_string().contains("null result") {
```
AsbjornOlling, line 289: Null results could be several things. For example, if llama.cpp doesn't recognize the architecture field, if the file is corrupted, or other things like that. I'd soften the error message for this a bit, and let it still include a text like "This can happen if you're trying to load a model that's bigger than the available memory" or similar.

AsbjornOlling (approved):

PR #534, opened 2026-05-18
fix: Add the `try` to swift Chat constructor

marek-hradil:
This is likely what is failing the pipelines currently. `Chat` now needs to be explicitely handled, cause it can fail.


AsbjornOlling:
Also fixed in #533
But there's something else broken...

PR #537, opened 2026-05-19
feat: Kokoro TTS

krebso:
## Description
Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

## Checklist:


nobodywho/crate-hashes.json
```diff
@@ -1,10 +1,10 @@
 {
-  "git+https://github.com/PyO3/pyo3?branch=main#0.28.3": "1srzibipk5z7miayz24n6sp2nf8vxshx18yazx0s16x4p4hb99a2",
```
krebso, line 2: Not sure if this is the correct resolution (why does hash change for the same revision), but otherwise the nix flake check fails.

nobodywho/core/Cargo.toml
```diff
@@ -46,6 +46,10 @@ libc = "0.2"
 
 [target.'cfg(not(target_os = "android"))'.dependencies]
```
AsbjornOlling, line 63: Is it correct that we're not building TTS for android?

nobodywho/core/src/tts/backends/kokoro.rs
```diff
@@ -0,0 +1,221 @@
...
+    language: String,
+    speed: f32,
+    /// Kokoro supports style for finite number of phonemes
+    max_input_phonemes: usize,
+}
+
+impl KokoroBackend {
+    pub fn new(
+        model_dir: &Path,
+        voice: String,
```
AsbjornOlling: Looks like this could be a `&str`?

nobodywho/core/src/tts/backends/kokoro.rs
```diff
@@ -0,0 +1,221 @@
...
+        // latter, so we do the wrap here.
+        // See https://github.com/hexgrad/kokoro/blob/main/kokoro/model.py#L130
+        let mut tokens: Vec<i64> = Vec::with_capacity(phoneme_ids.len() + 2);
+        tokens.push(0);
+        tokens.extend(phoneme_ids);
+        tokens.push(0);
+        let token_len = tokens.len();
+
+        let tokens = Tensor::from_array(([1usize, token_len], tokens))
+            .map_err(|e| TtsError::Synthesis(format!("tokens tensor: {e}")))?;
```
AsbjornOlling: For these .map_err cases, it may be cleaner to leverage thiserror's `#[from]` macro thingy. It would let us cast these implicitly with `?` and also keep the entire error stack, instead of flattening into a string here.

nobodywho/python/nobodywho.pyi
```diff
@@ -9,1143 +9,181 @@ T = typing.TypeVar('T', str, typing.Awaitable[str])  # Type variable for tool re
 
 @final
 class Audio:
-    """
```
AsbjornOlling, line 14: Looks like the typestubs file was regenerated using a version of pyo3 that doesn't support emitting docstrings.

This feature is the reason we were relying on pyo3 from git, and not crates.io

AsbjornOlling:
Looks like there's some build issues when running the full CI. Have a look here:

https://github.com/nobodywho-ooo/nobodywho/actions/runs/27766740483

The linux failure seems to just be missing some dependency.

Android looks a bit trickier. Seems like it normally fetches binaries at build-time, and we may need to build some ORT runtime from scratch since these aren't provided.

@krebso

AsbjornOlling (dismissed):
full-ci fails. Needs some build-time deps for linux and android.

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -0,0 +1,440 @@
...
+}
+
+impl Phonemizer {
+    fn new(
+        dialect: EspeakDialect,
+        espeak_lang: &str,
+        espeak_data_dir: Option<&Path>,
+    ) -> Result<Self, TtsError> {
+        let g2p = if matches!(dialect, EspeakDialect::EnGb | EspeakDialect::EnUs) {
+            use misaki_rs::{language::Language, G2P};
```
marek-hradil: top import

.github/workflows/build.yml
```diff
@@ -267,6 +267,15 @@ jobs:
...
 
+      # ORT's build.py uses `match` statements; needs Python 3.10+. Self-hosted mac
+      # runner's default python3 may be older. setup-python fails on self-hosted
+      # (writes to /Users/runner which doesn't exist), so install via brew.
+      - name: "Setup Python for ORT source build"
+        if: contains(matrix.target, 'visionos') || contains(matrix.target, 'watchos') || matrix.target == 'x86_64-apple-darwin'
+        run: |
+          brew install python@3.13
+          echo "/opt/homebrew/opt/python@3.13/libexec/bin" >> $GITHUB_PATH
+
```
marek-hradil, line 283: could we maybe move the ort related steps to a workflow which is separate and invoked by this workflow? also it would be nice to cache it, i reckon there will not be a need to rebuild it so often.

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -0,0 +1,440 @@
...
+            // Non-English: swap parens to angle quotes so espeak doesn't read
+            // `()` as language-switch markers. Mirrors misaki/espeak.py:89-106.
+            let espeak_input = text.replace('(', "«").replace(')', "»");
+            let raw_ipa = self
+                .translator
+                .text_to_ipa(&espeak_input)
+                .map_err(|e| TtsError::Synthesis(format!("espeak phonemization failed: {e}")))?;
+            let raw_ipa = raw_ipa.replace('«', "(").replace('»', ")");
+            Ok(espeak_ipa_to_misaki(&raw_ipa, self.dialect))
+        }
```
marek-hradil: split this function -> seems like its doing too many things at once and could use being splitted, also better for understanding the code.

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -0,0 +1,440 @@
...
+}
+
+/// Strip any stray ZWJ characters from misaki-rs output.
+fn zwj_to_single_char(s: &str) -> String {
+    s.replace('\u{200d}', "")
+}
+
+/// Extract bundled espeak-ng data on first use and build a Translator.
+/// Idempotent — re-running on an existing dir is safe.
+fn init_translator(language: &str, dir: Option<&Path>) -> Result<Translator, TtsError> {
```
marek-hradil: look if there is a already provided interface from the crate, so we dont have to extract it like this manually.

if not - check the location is always right

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -0,0 +1,440 @@
...
+    for (old, new) in [
+        ("aɪ", "I"),
+        ("aʊ", "W"),
+        ("ɔɪ", "Y"),
+        ("eɪ", "A"),
+        ("dʒ", "ʤ"),
+        ("tʃ", "ʧ"),
+        ("ɚ", "əɹ"),
+        ("əl", "ᵊl"),
+    ] {
```
marek-hradil: i would extract this to a constant at the top of a file.

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -0,0 +1,440 @@
...
+        }
+        EspeakDialect::NonEnglish => {
+            // misaki/espeak.py EspeakG2P.e2m extras beyond the shared set.
+            // GOAT `oʊ`→`O` applies for non-English too (the Romance languages
+            // espeak en-US gives produces oʊ for words like Portuguese GOAT;
+            // single-char keeps us consistent with the lexicon-trained model).
+            for (old, new) in [("oʊ", "O"), ("dz", "ʣ"), ("ts", "ʦ"), ("ss", "S")] {
+                s = s.replace(old, new);
+            }
+        }
```
marek-hradil: same here, this could be a map:

dialect -> replacements

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -0,0 +1,440 @@
...
+        ("r", "ɹ"),
+        ("x", "k"),
+        ("ç", "k"),
+        ("ɐ", "ə"),
+        ("ɬ", "l"),
+        ("ʲ", ""),
+        ("ɾ", "T"),
+        ("ʔ", "t"),
+        ("ʰ", ""), // espeak marks aspiration; misaki's lexicon never does
+    ] {
```
marek-hradil: same here

krebso:
[pipeline passes](https://github.com/nobodywho-ooo/nobodywho/actions/runs/27942268933/job/82719471444)

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -0,0 +1,440 @@
...
+            out.push('ᵊ');
+            out.push(chars[i]);
+            i += 2;
+        } else {
+            out.push(chars[i]);
+            i += 1;
+        }
+    }
+    out.replace('\u{0329}', "")
+}
```
marek-hradil: and then again split to a smaller functions.

what i think would be nice: there seems to be a lot of work associated with text -> espeak -> misaki conversions. maybe being able to represent it by a type would help the reader make sense out of it and scope it.

Something like `ESpeak::from_text(text)` then you work with the ESpeak struct until youre able to convert it with something like `Misaki::from_espeak(...)`.

_(my understanding of the problem is not great, but this illustrates my point and could be helpful with more TTS backends later on)_

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -0,0 +1,440 @@
...
+        let mut phoneme_ids: Vec<i64> = Vec::with_capacity(phonemes.len());
+        for ch in phonemes.chars() {
+            let mut buf = [0u8; 4];
+            let s = ch.encode_utf8(&mut buf);
+            if let Some(&id) = self.vocab.get(s) {
+                phoneme_ids.push(id);
+            }
+            // Unmapped IPA characters are dropped silently — same as upstream.
+            // https://github.com/hexgrad/kokoro/blob/main/kokoro/model.py#L128
+        }
```
marek-hradil: separate func candidate

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -0,0 +1,440 @@
...
+        let output = outputs[0].try_extract_tensor::<f32>()?;
+
+        let pcm = output.1.to_vec();
+        debug!(
+            phoneme_ids = token_len - 2,
+            pcm_samples = pcm.len(),
+            pcm_duration_s = pcm.len() as f32 / DEFAULT_SAMPLE_RATE as f32,
+            "Kokoro: done"
+        );
+        Ok((pcm, DEFAULT_SAMPLE_RATE))
```
marek-hradil: separate func candidate

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -0,0 +1,440 @@
...
+        .data()
+        .chunks_exact(4)
+        .map(|c| f32::from_le_bytes(c.try_into().unwrap()));
+    let voice_style: Vec<[f32; STYLE_DIM]> = floats
+        .collect::<Vec<f32>>()
+        .chunks_exact(STYLE_DIM)
+        .map(|row| row.try_into().expect("STYLE_DIM-sized chunk"))
+        .collect();
+    Ok((voice_style, rows - 1))
+}
```
marek-hradil: again, would be nice if the Voice was a struct, and this was sort of the init method.

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    let path = config_path.display();
+    let file = std::fs::File::open(config_path)
+        .map_err(|e| TtsError::Init(format!("kokoro: {path}: {e}")))?;
+    let Config { vocab } = serde_json::from_reader(file)
+        .map_err(|e| TtsError::Init(format!("kokoro: {path}: {e}")))?;
+    if vocab.is_empty() {
+        return Err(TtsError::Init(format!("kokoro: {path}: vocab is empty")));
+    }
```
marek-hradil: loading from a path repeats quite often.

could we maybe generalize somehow if we need to keep it?

marek-hradil (changes requested):
+ check the errors, ideally so they dont include the string description specified by the programmer, but are static only accepting params. `miette` can help you to make them nice for the user, if they are potentially user facing.

krebso:
@marek-hradil I will make the macos pipeline work in the meantime, all were good suggestions that I adressed.

duarteocarmo (approved):

nobodywho/core/src/lib.rs
```diff
@@ -2,13 +2,15 @@ pub mod chat;
 pub mod crossencoder;
 pub mod encoder;
 pub mod errors;
+pub(crate) mod huggingface;
```
marek-hradil:
```suggestion
pub mod huggingface;
```

marek-hradil (approved):

PR #539, opened 2026-05-20
Feat/rework docs

s194042:
Switch our docs from mkdocs to docusauraus. The main reason for this is to be able to do binding specific versioning of our docs. For now this is quite manual, but the workflow for during so is documented in the new readme.


docs/src/components/OpenInDropdown.tsx
```diff
@@ -0,0 +1,170 @@
...
+  markdown: '<svg viewBox="0 0 208 128" width="16" height="16" fill="none" stroke="currentColor" stroke-width="10"><rect x="5" y="5" width="198" height="118" rx="15" ry="15"/><path d="M30 98V30h20l20 25 20-25h20v68H90V59L70 84 50 59v39zM155 98V30h20v34h20l-30 34-30-34h20v34" fill="currentColor" stroke="none"/></svg>',
+};
+
+function Icon({name}: {name: keyof typeof ICONS}) {
+  return <span style={{display: 'inline-flex', alignItems: 'center', flexShrink: 0}} dangerouslySetInnerHTML={{__html: ICONS[name]}} />;
+}
+
+function deriveSourcePath(pathname: string): string {
+  const clean = pathname.replace(/\/$/, '') || '/docs';
+  const prefixMap: Record<string, string> = {
```
AsbjornOlling, line 25: Are these URLs correct? Opening the raw markdown file in the "Open-in" page seems to 404 for me

E.g. it will give this url https://docs.nobodywho.ooo/docs-python/index.md
Which I would expect maps to http://localhost:3000/docs-python/index.md
Which is 404

docs/migrate.py
AsbjornOlling, line 1: Is this file still needed?

docs/src/components/GitHubStars.tsx
AsbjornOlling, line 1: Is this file used? It seems like only `src/github-stars.js` is referenced.

docs/docusaurus.config.ts
```diff
@@ -0,0 +1,203 @@
...
+  baseUrl: '/',
+
+  organizationName: 'nobodywho-ooo',
+  projectName: 'nobodywho',
+
+  onBrokenLinks: 'throw',
+
+  markdown: {
+    hooks: {
+      onBrokenMarkdownImages: 'warn',
```
AsbjornOlling: maybe this should be `throw` like `onBrokenLinks`, to detect bad urls earlier

AsbjornOlling (approved):

PR #562, opened 2026-06-09
Refactor/SamplerBuilder typestate pattern and exposed the seed parameter

s194042:


AsbjornOlling:
Needs regeneration of flutter doctests

AsbjornOlling:
It would be good to also make seed configurable in godot bindings, to meet users' requests.

nobodywho/core/src/chat.rs
```diff
@@ -1662,14 +1662,16 @@ impl Worker<'_, ChatWorker> {
...
+                steps.insert(
+                    0,
+                    ShiftStep::Grammar {
                         trigger_on: tool_call_begin.clone(),
                         root: tool_grammar.root_name.to_string(),
                         grammar: tool_grammar.as_str().into(),
-                    })
+                    },
+                );
+                SamplerConfig::new(steps, self.extra.sampler_config.sample_step.clone())
```
AsbjornOlling: I think the seed variable might be reset here, when making a new sampler config with just the steps. It should also carry over the seed.

nobodywho/core/src/sampler.rs
```diff
@@ -272,14 +254,66 @@ impl SamplerConfig {
...
+
+    /// Appends a shift step to the end of the sampler chain.
+    pub fn shift(mut self, step: ShiftStep) -> Self {
+        self.steps.push(step);
+        self
+    }
+
+    /// Prepends a shift step to the beginning of the sampler chain.
+    /// This ensures the step is applied before any other shift steps.
+    pub fn prepend(mut self, step: ShiftStep) -> Self {
```
AsbjornOlling: I think this function is no longer used?
We could either change the `.insert(0, ...` calls to call prepend instead, or remove prepend.

AsbjornOlling (approved):

PR #563, opened 2026-06-09
fix(core): render LFM2.5 chat templates (rewrite `{% generation %}` to no-op)

### LINKED ISSUE #558: [BUG] LFM2.5-VL crash
**Describe the bug**
After loading, the model is crashing when being asked a question (both with text and image and text only).

Model [LFM2.5-VL Q8 page](https://huggingface.co/LiquidAI/LFM2.5-VL-450M-GGUF/tree/main)

Here is the error:
```
NobodyWhoGenerated.NobodyWhoError.Error(message: "Error creating response: Error syncing context and reading prompt: Could not render messages Template failed to render: syntax error: unknown statement generation (in <string>:67)\n    Diagnostic severity: error\n    Caused by: Error syncing context and reading prompt: Could not render messages Template failed to render: syntax error: unknown statement generation (in <string>:67)\n    Caused by: Could not render messages Template failed to render: syntax error: unknown statement generation (in <string>:67)\n    Caused by: Template failed to render: syntax error: unknown statement generation (in <string>:67)\n    Caused by: syntax error: unknown statement generation (in <string>:67)\n")
```

**To Reproduce**
Ask a question to the model

**Environment:**
 - OS: macOS
 - Nobodywho Version: 2.0.0 (Swift)
 - LLM Model: LFM2.5-VL

**Additional context**
- Can easily be reproduced with the swift starter example project (paste model/projection in Asset folder)
- Not crashing with [LFM2-VL](https://huggingface.co/LiquidAI/LFM2-VL-450M-GGUF/tree/main), text generation works but image ingestion gives the following result when I ask "analyse image": [0.42, 0.64, 0.47, 0.77]

gergelyvagujhelyi:
## Problem (#558)

LFM2.5 chat templates wrap the assistant turn in HuggingFace's training-only `{% generation %}` / `{% endgeneration %}` blocks. `minijinja` can't parse those and aborts with `unknown statement generation`, so **every** request on an LFM2.5 model (e.g. LFM2.5-VL) crashes before generating.

## Fix

Rewrite those tags to no-op `{% if true %}` / `{% endif %}` blocks in `ChatTemplate::new` before rendering. The tags are no-ops at inference, and `{% if true %}` always renders its body — so this is the minijinja-grammar equivalent of how llama.cpp's own jinja parser handles them (it maps `generation`/`endgeneration` to a noop statement). Whitespace-control markers are preserved, and `add_generation_prompt` is never matched.

Single file: `core/src/template.rs` (the `rewrite_generation_tags` pass + two focused tests).

## Verified
- Unit tests for the rewrite + a render through `ChatTemplate` with a generation block.
- End-to-end: LFM2.5-VL-450M now loads and chats; existing Llama-3 / DeepSeek / Qwen template tests still pass (the rewrite is a no-op for them).

LFM2 **tool calling** is a separate concern, proposed in #564.

Fixes #558

### DIFF
--- a/nobodywho/core/src/template.rs
+++ b/nobodywho/core/src/template.rs
@@ -31,7 +31,13 @@ static MINIJINJA_ENV: LazyLock<Environment> = LazyLock::new(|| {
     env
 });
 
-/// Rewrite training-only `{% generation %}`/`{% endgeneration %}` tags to no-op `{% if true %}`/`{% endif %}` that minijinja can parse.
+/// Some chat templates use `{% generation %}`/`{% endgeneration %}` tags,
+/// these are non-standard jinja syntax, and rely on a jinja extension used by hf transformers
+/// It looks like support for extensions like this isn't coming to minijinja anytime soon:
+///   https://github.com/mitsuhiko/minijinja/discussions/178
+///   https://github.com/mitsuhiko/minijinja/issues/200
+///   https://github.com/mitsuhiko/minijinja/issues/815
+/// For now our best bet is to replace these tags with a no-op.
 fn rewrite_generation_tags(template: &str) -> std::borrow::Cow<'_, str> {
     static GENERATION_TAG: LazyLock<Regex> = LazyLock::new(|| {
         Regex::new(r"\{%(-?)\s*(end)?generation\s*(-?)%\}")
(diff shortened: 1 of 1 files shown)


gergelyvagujhelyi:
Or I can make an upstream fix in minijinja

AsbjornOlling:
Seems like the generation tags are from a non-standard extension to jinja that's implemented in huggingface transformers.

Looking into minijinja upstream, it looks like there's some resistance to implementing custom blocks in minijinja.
Our best route for now is probably to do this preprocessing ourselves, like you did in this PR.

https://github.com/mitsuhiko/minijinja/issues/200
https://github.com/mitsuhiko/minijinja/discussions/178

AsbjornOlling (approved):

PR #564, opened 2026-06-09
feat(core): LFM2 tool calling

gergelyvagujhelyi:
## What

Adds tool calling for Liquid AI's **LFM2 / LFM2.5** models.

LFM2 emits tool calls as a Python-style call list:
`<|tool_call_start|>[get_weather(location="Paris")]<|tool_call_end|>`

New `Lfm2Handler` (`core/src/tool_calling/lfm2.rs`):
- a **GBNF grammar** constraining generation to that shape (required-param + comma handling), reusing the JSON value grammar per argument — nested values (objects/arrays) are fully supported;
- a nesting/quote-aware **parser** that extracts the calls (lenient — also accepts Python `True`/`False`/`None`).

Detection (`core/src/tool_calling/mod.rs`): template markers (`<|tool_call_start|>` / `<|tool_list_start|>` / `<|tool_response_start|>`) with an `lfm` arch/name fallback (needed: the LFM2.5 text-model templates carry no markers, and some GGUFs have hash-only `general.name`).

Also:
- adds an **LFM leg to the python CI tool-calling matrix**;
- fixes two typos in test tool descriptions (`dimenensions`, `heigth`) — small LFM models copy the misspelled key verbatim into their JSON arguments, which is how the typos were found.

## CI model

The LFM leg runs **LFM2.5-VL-1.6B Q8_0** from [`NobodyWho/LFM2.5-VL-1.6B-GGUF`](https://huggingface.co/NobodyWho/LFM2.5-VL-1.6B-GGUF) — LiquidAI's file with a single metadata change: the chat template additionally renders the `tool_calls` field of assistant messages (the upstream template renders only `message.content`, so re-rendered history would lose the model's own calls). With that template the model passes **all 14 tool-calling tests with no skips or test special-casing**. Q8_0 matters: the Q4_0 quant of the same model drops the `print()` in the python-tool test.

## Verified

- Full 14-test python suite green on LFM2.5-VL-1.6B Q8_0 (twice, deterministic).
- Tool round-trips (single, multi-call blocks, nested args: sets / tuples / nested lists / dicts) exercised across LFM2.5-350M, LFM2.5-1.2B-Instruct, LFM2.5-VL-450M, LFM2-2.6B and LFM2-8B-A1B (`lfm2moe`).
- No regression on the other CI legs (run locally with the exact CI models/flags): Qwen3 14/14, Ministral3 14/14, Gemma4 14/14, FunctionGemma 10P/4S (its usual format skips).

## Known limitations / follow-up material

- **Stock LiquidAI templates and multi-turn tool use**: 21 of LiquidAI's 24 tool-capable GGUF repos ship chat templates that render tool *definitions* and *responses* but not the assistant's `tool_calls` (their docs use a markup-in-content convention instead). With structured tool-call storage, assistant tool-call turns re-render as empty turns on those files: models may re-issue calls in multi-turn conversations (observed on 1.2B/2.6B), and on LFM2-8B-A1B this can run away until the context overflows. The durable fix is upstream template updates — the one-line template diff used for the CI model is ready to submit to LiquidAI. Only the LFM2.5-VL-450M(+Extract) and LFM2.5-8B-A1B templates render `tool_calls` natively.
- **Sampling**: LFM models are markedly more reliable at LiquidAI's recommended settings (temperature 0.1, top_k 50, repetition_penalty 1.05) than at generic defaults — without the repetition penalty, small/no-system-prompt chats on the 8B can spiral into degenerate token loops. Worth documenting for users.
- **Latent gbnf bug** (separate PR material, pre-existing): `json_schema_to_grammar` emits a reference to an undeclared `json-value` rule for `{}` / `true` / array-without-`items` param schemas, so llama.cpp rejects the whole grammar and every ask in that session fails. Also reachable via Qwen3.5 with bare-array params.
- GGUFs whose templates have no tool support at all (gen-1 LFM2-350M/700M/1.2B, gen-1 LFM2-VL) correctly refuse at init with `NoToolTemplate`.

Pairs with #563 (merged) for the `{% generation %}` templates.

### DIFF
--- a/.github/workflows/python_ci.yml
+++ b/.github/workflows/python_ci.yml
@@ -360,8 +360,8 @@ jobs:
             url: https://huggingface.co/unsloth/gemma-4-E2B-it-GGUF/resolve/main/gemma-4-E2B-it-Q4_0.gguf
             format: Gemma4
           - name: LFM2.5-VL 1.6B
-            file: LFM2.5-VL-1.6B-Q8_0.gguf
-            url: https://huggingface.co/NobodyWho/LFM2.5-VL-1.6B-GGUF/resolve/main/LFM2.5-VL-1.6B-Q8_0.gguf
+            file: LFM2.5-VL-1.6B-Q8_0-vendor-sampling.gguf
+            url: https://huggingface.co/NobodyWho/LFM2.5-VL-1.6B-GGUF/resolve/main/LFM2.5-VL-1.6B-Q8_0-vendor-sampling.gguf
             format: LFM2.5
 
     steps:
--- a/.github/workflows/python_ci.yml
+++ b/.github/workflows/python_ci.yml
@@ -359,6 +359,9 @@ jobs:
             file: gemma-4-E2B-it-Q4_0.gguf
             url: https://huggingface.co/unsloth/gemma-4-E2B-it-GGUF/resolve/main/gemma-4-E2B-it-Q4_0.gguf
             format: Gemma4
+          - name: LFM2.5 350M
+            url: https://huggingface.co/LiquidAI/LFM2.5-350M-GGUF/blob/main/LFM2.5-350M-Q4_0.gguf
+            format: LFM2.5
 
     steps:
       - uses: actions/checkout@v5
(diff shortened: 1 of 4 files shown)


gergelyvagujhelyi:
[lfm2.5-vl-chat-template.txt](https://github.com/user-attachments/files/28772069/lfm2.5-vl-chat-template.txt)

AsbjornOlling (approved):

PR #575, opened 2026-06-15
improved error messages for dart function parsing

jona605a:
fixed two tests that used invalid or impossible dart function types. 
added improved error messaging, including reminding users that all function parameters might be "required". 
added tests for each type of error message.

### DIFF
--- a/nobodywho/flutter/rust/src/lib.rs
+++ b/nobodywho/flutter/rust/src/lib.rs
@@ -610,7 +610,7 @@ pub fn new_python_tool(
 }
 
 /// Converts a Dart function runtimeType string directly to a JSON schema
-/// Example input: "({String a, int b}) => String" or "() => String"
+/// Example input: "({required String a, required int b}) => String" or "() => String"
 /// Returns a JSON schema for the function parameters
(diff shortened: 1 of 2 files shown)


nobodywho/flutter/rust/src/lib.rs
```diff
@@ -1130,20 +1149,6 @@ mod tests {
...
     }
 
-    #[test]
-    fn test_empty_params() {
-        let schema =
-            dart_function_type_to_json_schema("({}) => String", &std::collections::HashMap::new())
-                .unwrap();
-        let expected = serde_json::json!({
-            "type": "object",
-            "properties": {},
```
AsbjornOlling, line 1140: I think we should support functions with no parameters. Rather than removing this test, I think we should update the parsing logic to support functions with no parameters.

AsbjornOlling (changes requested):

nobodywho/flutter/rust/src/lib.rs
AsbjornOlling: nevermind. that test *is* invalid

AsbjornOlling (approved):

PR #579, opened 2026-06-16
feat: STT

marek-hradil:
Adds support for whisper STT.
```python
from nobodywho import STT

stt = STT("onnx-community/whisper-base")
text = stt.transcribe_file("sound.mp3").completed()
print(text) # wassup
```
Or you can choose to transcribe the buffer straight away, with:
```
stt.transcribe_pcm(samples, sample_rate=16000).completed()
```
Just make sure the buffer is i16 PCM format.
We also support streaming out (you can iterate over the response).


nobodywho/patches/symphonia-bundle-mp3/README.md
duarteocarmo, line 1: Wonder if we could simplify things a bit here by only supporting wav files? We could get rid of the whole symphonia bundle, and use hound which we already depend on? Thoughts?

krebso (approved):

PR #581, opened 2026-06-16
fix: Cleaning selfhosted mac runners

marek-hradil:
## Description

Manually clean self-hosted Mac CI, so that after interrupt it does not crash the next job.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -189,8 +189,6 @@ jobs:
   cargo-build-macos:
     if: inputs.full_ci
     runs-on: [self-hosted, macOS, ARM64]
-    env:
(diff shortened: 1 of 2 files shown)


s194042 (approved):

.github/workflows/build.yml
```diff
@@ -189,6 +189,8 @@ jobs:
   cargo-build-macos:
     if: inputs.full_ci
     runs-on: [self-hosted, macOS, ARM64]
+    env:
+      CARGO_TARGET_DIR: /tmp/nobodywho-${{ matrix.integration }}-${{ matrix.target }}-${{ matrix.profile }}
```
AsbjornOlling: If this target dir is persistent across runs, I'm not sure it'll solve the issue of cache poisoning in the target dir.

It will isolate it to a integration+target+profile+runner pair, making it way less frequent, but if a bad file ends up in that dir, it'll stay across runs.

PR #585, opened 2026-06-17
Added import index.js to wrapper.js to bypass lazy imports in release

s194042:


s194042:
The main fix is in wrapper.ts. 

So before we only initialized the rust code in index.tsx, which is only pulled in to wrapper.ts through reexports. This is an issue in release mode as the reexports are lazy in release mode. This means that if a user does not touch one of the reexported functions/structs (SamplerBuilder, getCachedModels etc.) then the rust lib is never initialized and nothing works. 

The fix here is to move all the responsibility of index.tsx into wrapper.ts and reexport from nobodywho-generated instead of nobodywho-generated -> index.tsx -> wrapper.ts. Since index.tsx is generated by uniffi, we leave the file commited to not cause issues/confusion but it is no longer relevant and this is documented in DEVELOPMENT.md.


Additionally, there are also some regenerated stuff and forgotten imports.

AsbjornOlling (approved):

PR #560, opened 2026-06-08
Bump llama-cpp-rs (by pointing to my branch)

AsbjornOlling:

PR #586, opened 2026-06-17
fix: scope android-static-stdcxx to Android so Linux cdylibs link

gergelyvagujhelyi:
Fixes the `nix-flake-check` (and any Linux `cdylib`) link failure on #560, without modifying that branch — this PR targets `bump-llama-cpp-for-mtp` so it can be merged straight in.

## Problem

Bumping to upstream `llama-cpp-rs` 0.1.150 changed the meaning of the `static-stdcxx` feature. On the old `marek-hradil` fork it was *"Only has an impact on Android"*; upstream now applies it on **Linux** too, where `build.rs` adds the compiler's static-archive dir to the link search path. That makes the dynamic `-lgomp` resolve to the non-PIC `libgomp.a`, which can't go into a shared object:

```
ld: libgomp.a(barrier.o): relocation R_X86_64_TPOFF32 against hidden symbol `gomp_tls_data' can not be used when making a shared object
```

`static-stdcxx` was switched on because nobodywho-core enables `android-static-stdcxx` (→ `llama-cpp-sys-2/static-stdcxx`) for **every** target, including Linux desktop. On the fork that was harmless; with upstream it breaks every Linux `cdylib` (flutter, godot, python).

## Fix

Scope `android-static-stdcxx` to Android only. This restores `main`'s dynamic-libstdc++ behavior on Linux (so `-lgomp` resolves to `libgomp.so` again) and keeps Android's self-contained static libc++. The MTP / `common` work is untouched.

- `core/Cargo.toml`: move `android-static-stdcxx` from the base + desktop `llama-cpp-2` deps into the `cfg(target_os = "android")` block.
- `Cargo.nix`: mirror the change in the generated file (the Nix build reads this, not `Cargo.toml`).

## Verified locally

- `cargo metadata --locked` passes — `Cargo.lock` unchanged.
- `cargo tree -e features`: Linux x86_64 → `llama-cpp-sys-2` no longer has `static-stdcxx`; `aarch64-linux-android` still does.
- `Cargo.nix` parses and the flake evaluates to a valid derivation.

The real confirmation is this PR's CI.

### DIFF
--- a/nobodywho/Cargo.lock
+++ b/nobodywho/Cargo.lock
@@ -2327,8 +2327,7 @@ dependencies = [
 [[package]]
 name = "llama-cpp-2"
 version = "0.1.150"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e82ec86fd73aaa7d8f4898cc4693410c217431506ba3237ea5c856127f49d84d"
+source = "git+https://github.com/nobodywho-ooo/llama-cpp-rs?branch=v0.1.150-watchos#2f5de95c2405a49350df5db11e113dd324cd639e"
 dependencies = [
  "encoding_rs",
  "enumflags2",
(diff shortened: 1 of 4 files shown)


gergelyvagujhelyi:
Asbjörn looked at it and he said merge

PR #588, opened 2026-06-18
refactor: Split inference from chat

marek-hradil:
This is only a refactor; every functionality should stay the same.
- `chat.rs` - stateful stuff around chat, so keeping the messages, shifting, etc.
- `inference.rs` - the unstateful things ("give me message list, i will give you back the new answer")

this should be first step towards continuous batching -> now there is more clear distinction between what is the chat (history, messages, context shifting etc.) and what is inference (template, tokenization, etc.) although is still not perfect (inference engine stll leaks `model`).


jona605a (approved):
Very good refactor to get started on working closer with the cont. batching.

PR #589, opened 2026-06-18
feat: Translation support

marek-hradil:
This is a support for translation gemma.
API:
```
translator = Translator("translategemma.gguf", source="en", target="da")

translator.translate("Wassup").completed()
```

or

```
for token in translator.translate("Wassup"):
   print(token)
```


AsbjornOlling:
I think this might be done in a simpler and more general way by extending the `Prompt` object with a variant that might support arbitrary keys.

E.g. 

```python
prompt = Prompt([
  Object(source="da", target="cz", text="Julemanden får travlt i år")
])
chat.ask(prompt)
```

or maybe just 

```python
prompt = Prompt([{"text": "Julemanden får travlt i år", "source": "da", "target": "cz"}])
```

Happy to discuss

AsbjornOlling:
As we spoke about IRL:

The role of the Translator class is not just to provide extra fields on the message structs, but also to make sure to clear chat history between invocations. Basically matching these two expectations of TranslateGemma.

Increasing the API surface area by an entire class just for TranslateGemma is a bit much.

We might solve it by implementing the chat-completion interface, similar to how many other llm libraries do it:

```
resp = chat.chat_completion([{
  "role": "user",
  "content": {
    "text": "Hold da kæft noget pæn kode",
    "target_language": "en",
    "source_langauge": "da",
  }
}])
```

PR #593, opened 2026-06-24
Remove flaky test

duarteocarmo:


nobodywho/python/tests/test_tool_calling.py
```diff
@@ -336,15 +336,12 @@ def test_tool_with_nested_list(chat):
 
     assert len(tool_calls) == 1
     assert tool_calls[0]["function"]["name"] == "add_list_of_vectors"
-    assert tool_calls[0]["function"]["arguments"]["list_of_vectors"] == [
-        [1, 2, 3],
-        [4, 5, 6],
-        [7, 8, 9],
```
marek-hradil, line 342: I dont think that at this point its useful to have test like this. If it is flaky, i would just remove it -> totally fine by me.

duarteocarmo: Agreed - removed

marek-hradil (approved):

PR #596, opened 2026-06-25
feat: add Supertonic TTS backend + python tts bindings

duarteocarmo:
- adds supertronic tts support
- adds python bindings as well

### DIFF
--- a/docs/docs-python/text-to-speech.md
+++ b/docs/docs-python/text-to-speech.md
@@ -1,0 +1,79 @@
+---
+title: Text-to-speech
+description: Generate audio from text with NobodyWho in Python.
+sidebar_position: 8
+---
+
+NobodyWho can synthesize text to WAV bytes with the `Tts` class.
+
+Two backends are available:
+
+- `kokoro`: lightweight 24 kHz speech synthesis.
+- `supertonic`: multi-stage ONNX speech synthesis with voice styles.
+
+Models are downloaded from Hugging Face on first use and cached locally.
+
+`backend` can be omitted for known official sources, including the examples below. For local directories or custom sources, pass `backend="kokoro"` or `backend="supertonic"`.
+
+## Kokoro
+
+Read more about Kokoro on its [official Hugging Face page](https://huggingface.co/hexgrad/Kokoro-82M). NobodyWho currently supports Kokoro language codes `en-us`, `en-gb`, `es`, `fr`, `it`, and `pt-br`.
+
+```python notest
+from pathlib import Path
+from nobodywho import Tts
+
+tts = Tts(
+    source="NobodyWho/Kokoro-82M",
+    voice="bf_emma",
+    language="en-gb",
+)
+
+wav = tts.synthesize("Hello from NobodyWho!")
+Path("out.wav").write_bytes(wav)
+```
+
+For Kokoro, set `voice` and `language` together. They must agree with the model's available voices.
+
+## Supertonic
+
+Read more about Supertonic on its [official Hugging Face page](https://huggingface.co/Supertone/supertonic-3), including supported languages and voice styles.
+
+```python notest
+from pathlib import Path
+from nobodywho import Tts
+
+tts = Tts(
+    source="Supertone/supertonic-3",
+)
+
+wav = tts.synthesize("Hello from NobodyWho!")
+Path("out.wav").write_bytes(wav)
+```
+
+By default, Supertonic uses `voice="M1"` and `language="en"`. The upstream model includes voice styles `M1`–`M5` and `F1`–`F5`.
+
+Most users can start with the defaults. Optional settings include:
+
+- `voice`: voice style, e.g. `M1` or `F1`. Defaults to `M1`.
+- `language`: input language code. Defaults to `en`.
+- `speed`: speech speed multiplier. Values above `1.0` are faster. Defaults to `1.05`.
+- `steps`: denoising steps. Higher can improve quality but is slower. Defaults to `8`.
+- `silence_duration`: silence inserted between long text chunks. Defaults to `0.3` seconds.
+
+## Async
+
+```python notest
+wav = await tts.synthesize_async("Hello!")
+```
+
+## Device selection
+
+By default, NobodyWho uses `device="auto"`. You can also choose `cpu` or `cuda`:
+
+```python notest
+tts = Tts(
+    source="Supertone/supertonic-3",
+    device="cpu",
+)
+```
--- a/nobodywho/core/Cargo.toml
+++ b/nobodywho/core/Cargo.toml
@@ -45,6 +45,11 @@ indicatif = "0.18"
 walkdir = "2.5"
 hound = "3"
 safetensors = "0.5"
+ndarray = "0.17"
+rand_distr = "0.5"
+unicode-normalization = "0.1"
+emojis = "0.9.0"
+unicode-segmentation = "1.13.3"
 
 [target.'cfg(target_os = "android")'.dependencies]
 libc = "0.2"
--- a/nobodywho/core/src/tts/backend.rs
+++ b/nobodywho/core/src/tts/backend.rs
@@ -1,6 +1,6 @@
 use crate::errors::TtsError;
 use crate::huggingface;
-use crate::tts::{kokoro, TtsConfig, TtsDevice};
+use crate::tts::{kokoro, supertonic, TtsConfig, TtsDevice};
 use std::time::Instant;
 use tracing::info;
 
@@ -33,7 +33,7 @@ pub(super) fn load_backend(
     match config {
         TtsConfig::Kokoro(config) => {
             let init_start = Instant::now();
-            let model_dir = huggingface::resolve(huggingface::parse(&config.source)?)?;
+            let model_dir = huggingface::resolve_model_dir(&config.source)?;
             let backend = kokoro::KokoroBackend::new(
                 &model_dir,
                 &config.voice,
@@ -42,6 +42,21 @@ pub(super) fn load_backend(
                 device,
             )?;
             info!(elapsed = ?init_start.elapsed(), "Initialized Kokoro TTS");
+            Ok(Box::new(backend))
+        }
+        TtsConfig::Supertonic(config) => {
+            let init_start = Instant::now();
+            let model_dir = huggingface::resolve_model_dir(&config.source)?;
+            let backend = supertonic::SupertonicBackend::new(
+                &model_dir,
+                &config.voice,
+                &config.language,
+                config.steps,
+                config.speed,
+                config.silence_duration,
+                device,
+            )?;
+            info!(elapsed = ?init_start.elapsed(), "Initialized Supertonic TTS");
             Ok(Box::new(backend))
         }
     }
--- a/nobodywho/core/src/tts/supertonic.rs
+++ b/nobodywho/core/src/tts/supertonic.rs
@@ -1,0 +1,661 @@
+use crate::errors::TtsError;
+use crate::tts::backend::TtsBackendImpl;
+use crate::tts::{ort_util, TtsDevice};
+use ndarray::{Array, Array3};
+use ort::session::Session;
+use ort::value::Tensor;
+use rand_distr::{Distribution, Normal};
+use serde::Deserialize;
+use std::fs::File;
+use std::io::BufReader;
+use std::path::{Path, PathBuf};
+use tracing::{debug, info};
+
+mod text;
+
+use text::{chunk_text, normalize_tts_text};
+
+const DEFAULT_SILENCE_DURATION: f32 = 0.3;
+const MAX_CHUNK_LENGTH: usize = 300;
+const MAX_CJK_CHUNK_LENGTH: usize = 120;
+const REQUIRED_ONNX_ASSETS: &[&str] = &[
+    "tts.json",
+    "unicode_indexer.json",
+    "duration_predictor.onnx",
+    "text_encoder.onnx",
+    "vector_estimator.onnx",
+    "vocoder.onnx",
+];
+const SUPPORTED_LANGS: &[&str] = &[
+    "en", "ko", "ja", "ar", "bg", "cs", "da", "de", "el", "es", "et", "fi", "fr", "hi", "hr", "hu",
+    "id", "it", "lt", "lv", "nl", "pl", "pt", "ro", "ru", "sk", "sl", "sv", "tr", "uk", "vi", "na",
+];
+
+pub(in crate::tts) struct SupertonicBackend {
+    config: SupertonicTtsJsonConfig,
+    text_processor: UnicodeProcessor,
+    duration_predictor: Session,
+    text_encoder: Session,
+    vector_estimator: Session,
+    vocoder: Session,
+    sample_rate: i32,
+    style: Style,
+    language: SupertonicLanguage,
+    settings: SupertonicSettings,
+}
+
+struct SupertonicLanguage(String);
+
+impl SupertonicLanguage {
+    fn new(language: &str) -> Result<Self, TtsError> {
+        if SUPPORTED_LANGS.contains(&language) {
+            Ok(Self(language.to_string()))
+        } else {
+            Err(TtsError::UnsupportedLanguage {
+                language: language.into(),
+                supported: SUPPORTED_LANGS.join(", "),
+            })
+        }
+    }
+
+    fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    fn uses_short_chunks(&self) -> bool {
+        self.0 == "ko" || self.0 == "ja"
+    }
+
+    fn preprocess_text(&self, text: &str) -> String {
+        let text = normalize_tts_text(text);
+        format!("<{}>{text}</{}>", self.0, self.0)
+    }
+}
+
+struct SupertonicSettings {
+    steps: usize,
+    speed: f32,
+    silence_duration: f32,
+}
+
+impl SupertonicSettings {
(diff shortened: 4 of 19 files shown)


nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+const MAX_CJK_CHUNK_LENGTH: usize = 120;
+const REQUIRED_ONNX_ASSETS: &[&str] = &[
+    "tts.json",
+    "unicode_indexer.json",
+    "duration_predictor.onnx",
+    "text_encoder.onnx",
+    "vector_estimator.onnx",
+    "vocoder.onnx",
+];
+const SUPPORTED_LANGS: &[&str] = &[
```
marek-hradil, line 29: is there a way of providing unified language interface with kokoro? so the language enum does not change with changing the backend?

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+        voice: &str,
+        language: &str,
+        steps: usize,
+        speed: f32,
+        silence_duration: f32,
+        device: TtsDevice,
+    ) -> Result<Self, TtsError> {
+        validate_language(language)?;
+        validate_config(steps, speed, silence_duration)?;
+        ensure_assets_exist(model_dir, voice)?;
```
marek-hradil: Optional as this is more of my preference: I think these _throw if you see something wrong_ procedures can get very quickly very ugly.

What I think is more maintainable solution is to replace them with *constructors* which give you back the object if the data are valid. Why is that better? The ownership of the object itself then gives you the guarantee its right, you dont have to *call again* the validate_procedure on each new place where you intend to use the object.

If this is not clear, text me I can explain further.

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+}
+
+impl SupertonicBackend {
+    pub fn new(
+        model_dir: &Path,
+        voice: &str,
+        language: &str,
+        steps: usize,
+        speed: f32,
+        silence_duration: f32,
```
marek-hradil, line 140: is there a good reason to modify these? would it make sense to let the user use some presets instead of having control directly over the numbers?

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+struct VoiceStyleData {
+    style_ttl: StyleComponent,
+    style_dp: StyleComponent,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+struct StyleComponent {
+    data: Vec<Vec<Vec<f32>>>,
+    dims: Vec<usize>,
+}
```
marek-hradil, line 248: Lots of different configs here. Do we need to have them nested like this? Can't we simplify the structure to make the handling then easier? Or does this copy onnx structure?

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+
+impl UnicodeProcessor {
+    fn new(path: &Path) -> Result<Self, TtsError> {
+        Ok(Self {
+            indexer: read_json(path)?,
+        })
+    }
+
+    fn call(
```
marek-hradil: some more meaningful name please

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+
+        for text in &processed_texts {
+            let mut row = vec![0i64; max_len];
+            for (index, unicode_value) in text.chars().map(|c| c as usize).enumerate() {
+                row[index] = self.indexer.get(unicode_value).copied().unwrap_or(-1);
+            }
+            text_ids.push(row);
+        }
+
+        Ok((text_ids, length_to_mask(&lengths, max_len)))
```
marek-hradil: Youre returning quite complex structure, which you then convert into a tensors. Would be good if this was returning the tensor right away, as you seem to need to do the tensor conversion always anyways.

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+        let max_len = lengths.iter().copied().max().unwrap_or(0);
+        let mut text_ids = Vec::with_capacity(processed_texts.len());
+
+        for text in &processed_texts {
+            let mut row = vec![0i64; max_len];
+            for (index, unicode_value) in text.chars().map(|c| c as usize).enumerate() {
+                row[index] = self.indexer.get(unicode_value).copied().unwrap_or(-1);
+            }
+            text_ids.push(row);
+        }
```
marek-hradil: split this into a named method.

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+
+impl Style {
+    fn load(path: &Path) -> Result<Self, TtsError> {
+        let data: VoiceStyleData = read_json(path)?;
+        let ttl = component_to_array(&data.style_ttl, path, "style_ttl")?;
+        let dp = component_to_array(&data.style_dp, path, "style_dp")?;
+        Ok(Self { ttl, dp })
+    }
+}
+
```
marek-hradil, line 271: feels little weird being able to load Style, but then have "no way" of using this. Later as with UnicodeProcessor, you convert it into the tensor. Cant we have atleast a method for that, so its clear this is the intended usage?

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+        speed: f32,
+        silence_duration: f32,
+    ) -> Result<Vec<f32>, TtsError> {
+        let max_len = if uses_short_chunks(language) {
+            MAX_CJK_CHUNK_LENGTH
+        } else {
+            MAX_CHUNK_LENGTH
+        };
+        let chunks = chunk_text(text, max_len);
+        let mut wav = Vec::new();
```
marek-hradil: the whole chunking logic which begins here is quite complex, takes a lots of this file and seems its not necesarilly specific only to supertonic. Can we move it to a separate file and there make clear what is the entrypoint (probably chunk_text) and what are methods used only within?

Generally: It would be great to encapsulate it and move it away from this file.

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+            )?;
+            let wav_len = (self.sample_rate as f32 * duration[0]) as usize;
+            if index > 0 {
+                wav.extend(std::iter::repeat_n(
+                    0.0,
+                    (silence_duration * self.sample_rate as f32) as usize,
+                ));
+            }
+            wav.extend_from_slice(&chunk_wav[..wav_len.min(chunk_wav.len())]);
+        }
```
marek-hradil, line 339: good candidate for a method, so its more clear what youre doing here at a first sight (took me a while).

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+
+        Ok(wav)
+    }
+
+    fn infer(
```
marek-hradil: split this method and associate the underlying functions with the objects they modify. this is *too long*

nobodywho/core/src/tts/supertonic.rs
marek-hradil: Edit: same thing with the text preprocessing. Isnt this something kokoro could take advantage of?
Similary, normalization is not specific to supertonic engine -> my intuition would be it lives in another file.

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,857 @@
...
+    if voice_path.exists() {
+        return Ok(());
+    }
+
+    let available = list_voices(model_dir);
+    Err(TtsError::MissingVoice {
+        voice: voice.to_string(),
+        available,
+    })
+}
```
marek-hradil: Last general pattern: both kokoro and supertonic tend to have lots of logic associated with loading stuff from onnx.

*Optional*: Could you just investigate if some of it couldnt be extracted to something like `onnx.rs` where this would be handled?
It feels weird having to read lots of loading and checking files, when interested about how does the supertonic backend work.

marek-hradil (changes requested):

marek-hradil (changes requested):
Hey! Thanks for the changes, looks great!

I have mostly comments around the supertonic.rs, where I would like some things to be more general. Hope they make sense & feel reasonable.

nobodywho/core/src/tts/supertonic.rs
duarteocarmo: It just parses this: https://huggingface.co/Supertone/supertonic-3/blob/main/onnx/tts.json

I ignore the fields we don't need (serde does that by default) 
I renamed them so they are a bit better.

nobodywho/core/src/tts/supertonic.rs
duarteocarmo: Thought about this a bit. But think is a bit dangerous. 

Qwen tts supports "english", others "en", others nothing. Some "portuguese", others "pt-pt", "pt-br". I would prefer each backend validates the language and keep the api as a string. But let me know if you disagree.

nobodywho/core/src/tts/supertonic.rs
duarteocarmo: In the python bindings (and hopefully others) the users can completely omit these. We already have some defaults: 

```
 - steps = 8
 - speed = 1.05
 - silence_duration = 0.3
```

nobodywho/core/src/tts/supertonic.rs
duarteocarmo: Done, added some constructors - coming from Python I like the assert style - but point taken!

nobodywho/core/src/tts/supertonic.rs
duarteocarmo: I agree, the only thing I see common for now is `load_session` - which is already shared. The languages are a bit different, ad the text processing is also quite different. I just think they are different architectures. But leaving this open in case any ideas come up :)

nobodywho/core/src/tts/supertonic.rs
duarteocarmo: Moved the logic to another file. I think they are both specific to supertronic - and that's clear from the structure.

duarteocarmo:
@marek-hradil  went through your notes - I believe I addressed all. lmk

docs/docs-python/text-to-speech.md
```diff
@@ -0,0 +1,80 @@
...
+title: Text-to-speech
+description: Generate audio from text with NobodyWho in Python.
+sidebar_position: 8
+---
+
+NobodyWho can synthesize text to WAV bytes with the `Tts` class.
+
+Two backends are available:
+
+- `kokoro`: lightweight 24 kHz speech synthesis. This is the default when `backend` is not provided.
```
krebso: **suggestion**: I would make the backend mandatory to specify.

nobodywho/core/src/tts/backend.rs
```diff
@@ -44,9 +45,29 @@ pub(super) fn load_backend(
...
+            Ok(Box::new(backend))
+        }
     }
 }
 
+fn resolve_model_dir(source: &str) -> Result<PathBuf, TtsError> {
+    let source = huggingface::parse(source)?;
+    Ok(huggingface::resolve(source)?)
+}
+
```
krebso: Can we extract this to `huggingface.rs`?

nobodywho/core/Cargo.toml
```diff
@@ -45,6 +45,9 @@ indicatif = "0.18"
 walkdir = "2.5"
 hound = "3"
 safetensors = "0.5"
+ndarray = "0.17"
```
krebso, line 48: we are already using `safetensors`, can we use it instead?

nobodywho/core/src/tts/supertonic.rs
```diff
@@ -0,0 +1,728 @@
...
+    "vector_estimator.onnx",
+    "vocoder.onnx",
+];
+const SUPPORTED_LANGS: &[&str] = &[
+    "en", "ko", "ja", "ar", "bg", "cs", "da", "de", "el", "es", "et", "fi", "fr", "hi", "hr", "hu",
+    "id", "it", "lt", "lv", "nl", "pl", "pt", "ro", "ru", "sk", "sl", "sv", "tr", "uk", "vi", "na",
+];
+
+pub(in crate::tts) struct SupertonicBackend {
+    text_to_speech: TextToSpeech,
```
krebso: **nitpick**: I would prefer this struct to be inlined, adds one more step of indirection and then impl of the backend just wraps the method, and sometimes add functionality (`synthesize_raw`) that is probably needed, and the underlying method will return a wrong result without it.

nobodywho/core/src/tts/supertonic/text.rs
```diff
@@ -0,0 +1,221 @@
...
+    let text = normalize_tts_spacing(&text);
+    ensure_terminal_punctuation(&text)
+}
+
+fn remove_emoji(text: &str) -> String {
+    Regex::new(r"[\x{1F600}-\x{1F64F}\x{1F300}-\x{1F5FF}\x{1F680}-\x{1F6FF}\x{1F700}-\x{1F77F}\x{1F780}-\x{1F7FF}\x{1F800}-\x{1F8FF}\x{1F900}-\x{1F9FF}\x{1FA00}-\x{1FA6F}\x{1FA70}-\x{1FAFF}\x{2600}-\x{26FF}\x{2700}-\x{27BF}\x{1F1E6}-\x{1F1FF}]+")
+        .unwrap()
+        .replace_all(text, "")
+        .to_string()
+}
```
krebso, line 17: **question**: I think here adding the deps actually might be worth: https://stackoverflow.com/a/76909909

nobodywho/core/src/tts/supertonic/text.rs
```diff
@@ -0,0 +1,221 @@
...
+            | '】'
+            | '〉'
+            | '》'
+            | '›'
+            | '»'
+    )
+}
+
+pub(super) fn chunk_text(text: &str, max_len: usize) -> Vec<String> {
+    let text = text.trim();
```
krebso, line 87: **question**: lot of `text.trim()`'s flying around. Can we do it on entry, and then make sure that the operations we do do not break the invariant?

nobodywho/core/src/tts/supertonic/text.rs
```diff
@@ -0,0 +1,221 @@
...
+    fn chunk_text_keeps_abbreviations_together() {
+        assert_eq!(
+            split_sentences("Dr. Smith arrived. Hello there."),
+            vec![
+                "Dr. Smith arrived. ".to_string(),
+                "Hello there.".to_string()
+            ]
+        );
+    }
+}
```
krebso, line 267: **question**: Can we add tests for all the sentence/chunk splitting behavior?

nobodywho/core/src/tts/kokoro.rs
```diff
@@ -548,15 +548,34 @@ impl KokoroVoice {
...
     pub speed: f32,
 }
 
 impl KokoroConfig {
+    /// Create a config with defaults for the given `source`.
+    ///
+    /// `source` is a HuggingFace repo id (`owner/repo`) or a local model directory.
     pub fn new(source: impl AsRef<str>) -> Self {
         Self {
             source: source.as_ref().to_string(),
```
krebso: **suggestion**: not sure if these are new comments or relic of some rebase, but I think they are redundant here.

nobodywho/python/src/lib.rs
```diff
@@ -204,6 +204,136 @@ impl<'py> ModelOrPath<'py> {
...
+        "cuda" => Ok(nobodywho::tts::TtsDevice::Cuda),
+        _ => Err(pyo3::exceptions::PyValueError::new_err(
+            "device must be one of 'auto', 'cpu', or 'cuda'",
+        )),
+    }
+}
+
+fn build_tts_config(
+    source: std::path::PathBuf,
+    backend: &str,
```
krebso: **question**: I was thinking, if we could infer the backend either from the `source`, or the files the source dir actually contains. Since we are already enforcing strict naming, I think we could dynamically try to load the backends and stop on the first one that succeeds. wdyt?

krebso (changes requested):

nobodywho/core/Cargo.toml
duarteocarmo: I think safetensors is good for the file format stuff, but for the matrix multiplication in multiple dimensions not so much? 

Example: 

 ```rust
   noisy_latent[[batch, dim, time]] *= latent_mask[[batch, 0, time]];
 ```

 vs.

 ```rust
   let latent_index = batch * latent_dim * latent_len + dim * latent_len + time;
   let mask_index = batch * latent_len + time;
   noisy_latent[latent_index] *= latent_mask[mask_index];
 ```
The code would become a vast bit more noisy.

docs/docs-python/text-to-speech.md
duarteocarmo: Made it mandatory - I did it like that since @marek-hradil  mentioned the more default the better.

nobodywho/python/src/lib.rs
duarteocarmo: Good idea - added.

nobodywho/core/Cargo.toml
krebso: sure, i did not realize that we use them also for tensor ops.

krebso (approved):

krebso (approved):

marek-hradil (approved):

PR #601, opened 2026-07-01
TTS Bindings

duarteocarmo:


docs/docs-flutter/text-to-speech.md
```diff
@@ -0,0 +1,90 @@
...
+- `silenceDuration`: silence inserted between long text chunks. Defaults to `0.3` seconds.
+
+## Local model folder format
+
+When `source` is a local directory, point it at the top-level model folder and pass the matching `backend`.
+
+Use the Hugging Face file browsers as the reference layouts:
+
+- Kokoro: [`NobodyWho/Kokoro-82M`](https://huggingface.co/NobodyWho/Kokoro-82M/tree/main)
+- Supertonic: [`Supertone/supertonic-3`](https://huggingface.co/Supertone/supertonic-3/tree/main)
```
pielouNW: https://huggingface.co/Supertone/supertonic-3/tree/main/onnx

docs/docs-flutter/index.md
```diff
@@ -44,7 +44,7 @@ print(msg); // Yes, indeed, water is wet!
 
 This is a super simple example, but we believe that examples which do simple things, should be simple!
 
-To get a full overview of the functionality provided by NobodyWho, simply keep reading. You can also have a look at our [flutter starter app repository](https://github.com/nobodywho-ooo/flutter-starter-example).
+To get a full overview of the functionality provided by NobodyWho, simply keep reading. For local audio generation, see [Text-to-speech](./text-to-speech). You can also have a look at our [flutter starter app repository](https://github.com/nobodywho-ooo/flutter-starter-example).
```
marek-hradil: this does not make sense to have here, if the other features are not mentioned.

docs/docs-flutter/text-to-speech.md
```diff
@@ -0,0 +1,94 @@
+---
+title: Text-to-speech
+description: Generate WAV audio from text with NobodyWho in Flutter.
+sidebar_position: 6
+---
+
```
marek-hradil, line 6: sorry, but this feels messy - and thats **okay**, we dont have a clear specification of how the docs should look like.

_still_, it feels like it brings the complex details first, instead of focusing on the simple use-case first.

what i would do with it:

1. start with the context (what is this functionality for?), can be super short
2. introduce the most easy example, probably something like you have below:
```
import 'dart:io';
import 'package:nobodywho/nobodywho.dart' as nobodywho;

// ... requires a downloaded or local TTS model.
final tts = await nobodywho.Tts.newInstance(
  source: 'NobodyWho/Kokoro-82M',
  voice: 'bf_emma',
  language: 'en-gb',
);

final wav = await tts.synthesize(text: 'Hello from NobodyWho!');
await File('out.wav').writeAsBytes(wav);
```
3. then follow with sections about language, voices, etc.

keep in mind that our biggest value (as we are mostly wrapping existing implementations) is **simplifications and providing good defaults** - most people are not AI engineers and dont want to be. for this reason, i would try to think about how can we abstract away (or make more comfy) denoising `steps`.

if youre feeling something from this is wrong or should be done in other fashion lets discuss. i know this is not yet clearly specified, but i feel having a clean and simple docs helps.

docs/docs-flutter/text-to-speech.md
```diff
@@ -0,0 +1,94 @@
...
+- Kokoro: [`NobodyWho/Kokoro-82M`](https://huggingface.co/NobodyWho/Kokoro-82M/tree/main)
+- Supertonic: [`Supertone/supertonic-3`](https://huggingface.co/Supertone/supertonic-3/tree/main)
+
+For Supertonic, that top-level folder must include both the `onnx/` and `voice_styles/` directories. Download the model files with the same relative paths, then pass that folder as `source`.
+
+## Sync
+
+```dart
+// ... create tts as shown above.
+final wav = tts.synthesizeSync(text: 'Hello!');
```
marek-hradil: if im not mistaken, for flutter we dont provide sync functions as everything is mostly async. check and if its true, then remove this.

docs/docs-flutter/text-to-speech.md
```diff
@@ -0,0 +1,94 @@
...
+
+By default, NobodyWho uses `device: 'auto'`. You can also choose `cpu` or `cuda`:
+
+```dart
+// ... requires a downloaded or local TTS model.
+final tts = await nobodywho.Tts.newInstance(
+  source: 'Supertone/supertonic-3',
+  device: 'cpu',
+);
+```
```
marek-hradil: 1. In order to be unified with `Chat`, I would drop `device` and keep it as `useGpu` bool.
2. If you feel `device` is a better way to go, we can discuss and then change it on both places.

docs/docs-flutter/text-to-speech.md
```diff
@@ -0,0 +1,94 @@
...
+## Supertonic
+
+Read more about Supertonic on its [official Hugging Face page](https://huggingface.co/Supertone/supertonic-3), including supported languages and voice styles.
+
+```dart
+import 'dart:io';
+import 'package:nobodywho/nobodywho.dart' as nobodywho;
+
+// ... requires a downloaded or local TTS model.
+final tts = await nobodywho.Tts.newInstance(
```
marek-hradil: why not `.load` as for the other flutter classes?

nobodywho/godot/src/lib.rs
```diff
@@ -47,6 +47,76 @@ fn resolve_godot_path(path: &GString) -> String {
     }
 }
 
+fn parse_tts_backend(backend: String) -> Result<Option<nobodywho::tts::TtsBackendKind>, GString> {
```
marek-hradil, line 50: _thought for discussion_ (this is not blocking and can be merged without addressing this).

the parsing is duplicated over all the integrations.

pros: i feel like this is the right place to do the conversion from strings to enums

cons: i am little worried that more use cases like this will appear and then it will (1) bloat the codebase with lots of same code for every binding and (2) the error messages will be hard to maintain as they are not part of the core "error registry".

is there a way to approach this?

marek-hradil (changes requested):
Good job, but more human input is needed on the docs side of things. Feel free to discuss.

docs/docs-flutter/text-to-speech.md
duarteocarmo: Nice - too another stab - lmk what you think.

docs/docs-flutter/text-to-speech.md
duarteocarmo: Updated.

nobodywho/godot/src/lib.rs
duarteocarmo: I thought about this when we merged the python bindings. Since we ultimately decided to keep them at the python layer - I think it should sit at that layer for every other provider. We can probably improve this in the future - but would not see it as a blocker per se.

docs/docs-flutter/text-to-speech.md
duarteocarmo: True - nice catch.

duarteocarmo:
@marek-hradil  - thanks for the notes. Agree with most. Took another stab at the docs - also added a note about Apple acceleration for the Swift side of things. Lmk what you think.

docs/docs-flutter/text-to-speech.md
```diff
@@ -0,0 +1,118 @@
...
+  language: 'en-gb',
+);
+
+// Generate WAV bytes for this sentence.
+final wav = await tts.synthesize(text: 'Hello from NobodyWho!');
+// Save the audio to a file.
+await File('out.wav').writeAsBytes(wav);
+```
+
+That was a lot. No need to panic. Start with `source`: it tells NobodyWho which TTS model to load. More on that in the next section.
```
pielouNW: Sounds weird. Can be deleted imo

docs/docs-flutter/text-to-speech.md
```diff
@@ -0,0 +1,118 @@
...
+
+// ... after NobodyWho.init().
+final tts = await nobodywho.Tts.load(
+  // Hugging Face repo ID or local folder with the model files.
+  source: 'NobodyWho/Kokoro-82M',
+  // Voice to use from the model.
+  voice: 'bf_emma',
+  // Language code for the input text.
+  language: 'en-gb',
+);
```
pielouNW, line 22:
```
final tts = await nobodywho.Tts.load(
  source: 'NobodyWho/Kokoro-82M', // Hugging Face repo ID or local folder with the model files.
  voice: 'bf_emma', // Voice to use from the model.
  language: 'en-gb', // Language code for the input text.
);
```

marek-hradil (approved):

PR #604, opened 2026-07-03
Improve PR CI performance

duarteocarmo:
faster CI, automatic workflow cancellation + caching.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,6 +12,10 @@ on:
         default: true
         description: 'Run full CI suite (all platforms, all tests)'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
 permissions:
   contents: write
   pages: write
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -51,6 +55,7 @@ jobs:
     uses: ./.github/workflows/test.yml
     with:
       full_ci: ${{ github.event_name == 'push' || (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'full-ci')) || (github.event_name == 'workflow_dispatch' && inputs.full_ci) }}
+      run_nix: ${{ github.event_name == 'push' || (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'full-ci')) || (github.event_name == 'workflow_dispatch' && inputs.full_ci) }}
```
s194042, line 58: What is the reason for disabling the nix-flake check? In my experience it is usually pretty fast.

duarteocarmo: Just that I don't think we need it for PRs? It still runs on main, and on full-ci - so we will know if it fails. Just think not a critical path?

AsbjornOlling (approved):

PR #606, opened 2026-07-04
feat: Whisper quantization + downloading improvements

marek-hradil:
fix: @duarteocarmo discovered that when downloading onnx repo, the implementation downloads much more than needed. that is now fixed.

feat: along with that, we now support different quantizations and let the user choose them

fix: lastly, there were some remaining issues with .wav codecs. seems like for this reason we need symphonia after all (even if we decided to drop the mp3 support). 

refactor: huggingface.rs is now more split between the onnx path and the gguf path. both use same cache, which is now better at persisting when the download fails.


nobodywho/core/src/stt/backends/whisper.rs
```diff
@@ -29,24 +29,71 @@ const MAX_NEW_TOKENS: usize = 448;
...
+pub const DEFAULT_QUANTIZATION: &str = "default";
+
+/// Map a user-supplied quantization name to the ONNX filename suffix used by
+/// `onnx-community/whisper-*` repos, e.g. `"fp16"` -> `"_fp16"`.
+///
+/// `onnx-community/whisper-*` repos ship the same encoder and decoder graph
+/// exported at several precisions (see [`WhisperBackend`] docs for why only
+/// the *merged* decoder variant is used, never the `_with_past` ones).
+/// Picking one avoids downloading every variant just to use one of them.
+fn quantization_suffix(quantization: &str) -> Result<&'static str, SttError> {
```
AsbjornOlling, line 50: If we only support a short list of quantization levels, it should maybe be an enum instead of a &str?

We can always make the from/into stuff to easily coerce from a string passed by the user.

nobodywho/core/src/stt/backends/whisper.rs
```diff
@@ -29,24 +29,71 @@ const MAX_NEW_TOKENS: usize = 448;
...
+/// ships. Downloading only these avoids pulling down gigabytes of unused
+/// variants and the never-used `_with_past` graphs.
+pub(in crate::stt) fn required_files(quantization: &str) -> Result<Vec<String>, SttError> {
+    let suffix = quantization_suffix(quantization)?;
+    Ok(vec![
+        "config.json".into(),
+        "generation_config.json".into(),
+        "tokenizer.json".into(),
+        format!("onnx/encoder_model{suffix}.onnx"),
+        format!("onnx/decoder_model_merged{suffix}.onnx"),
```
AsbjornOlling, line 102: How strong are these filename conventions? I fear that we're setting ourselves up to only download whisper models from the hf onnx-community org.

Alternately we could ask the user to provide a complete path to the model file, or (if we can list files in the repo), filter for files that *contain* the name of the quantization level, instead of expecting a precise filename.

But if the convention is really strong, it's probably fine.

nobodywho/core/src/stt/backends/whisper.rs
```diff
@@ -29,24 +29,71 @@ const MAX_NEW_TOKENS: usize = 448;
 // Public types
 // ---------------------------------------------------------------------------
 
+/// Default value of [`WhisperConfig::quantization`] — the unsuffixed fp32 ONNX variant.
+pub const DEFAULT_QUANTIZATION: &str = "default";
```
AsbjornOlling: If quantization levels were an enum, this could be the default value of that type.

nobodywho/core/src/huggingface.rs
```diff
@@ -378,148 +440,269 @@ pub(crate) fn download_model_from_hf(
...
+/// directories win; otherwise `source` is treated as a HuggingFace Hub repo
+/// ID (`owner/repo`) and downloaded, narrowed to `required_files` (an empty
+/// slice downloads the entire repo unfiltered).
+pub(crate) fn resolve_model_dir(
+    source: &str,
+    required_files: &[String],
+) -> Result<PathBuf, HuggingFaceError> {
+    let cache = ModelCache::open()?;
+    OnnxSource::parse(source, required_files.to_vec())?
+        .resolve(&cache, &default_progress_callback())
```
AsbjornOlling: Looks like this codepath removes the users' ability to pass in a custom download-progress callback.

nobodywho/core/src/stt/backends/whisper.rs
AsbjornOlling: Also: is `fp32` really the default quantization we want to recommend? For text generation models we usually recommend q4. Is whisper more heavily degraded with quantization?

AsbjornOlling (changes requested):

nobodywho/core/src/stt/backends/whisper.rs
marek-hradil: yes. but i would like to do this in a bigger batch also with the languages. i feel like we should be always accepting enums at least with the rust crate.

let me postpone this and i will then refactor it also with the other enums which are now strings.

nobodywho/core/src/stt/backends/whisper.rs
marek-hradil: yup, had a same problem with this, but onnx repos seem to be opposite of standardized, so filtering out only things you need is hard.

@duarteocarmo had some strong opinions on this, maybe he has a solution.

duarteocarmo: I think it would be ok to standardize accross the onnx community repos. And, at least of the onnx versions of whisper - it should be pretty consistent with the files.

nobodywho/core/src/stt/backends/whisper.rs
marek-hradil: so far changed for q4. will add the enum later.

nobodywho/core/src/huggingface.rs
marek-hradil: The problem here is that for TTS/STT we dont have the construction with `download_file` where we could pass in the `on_download_progress ` as we dont have the explicit `Model` object.

Im not sure how to solve this nicely yet. A nice way could be maybe:
```python
model = download_file("...", on_download_progress=...)

STT(model)
```
but im not sure if this is technically possible

AsbjornOlling: I'll add a ticket and we can do it later :+1:

AsbjornOlling (approved):

PR #612, opened 2026-07-07
feat: Voice Activity Detection

marek-hradil:
`core/` implementation for voice activity detection. was quite fast so ive done it.

so far works like this:

```rust
    let vad = Vad::new(VadConfig {
        sample_rate,
        source: "onnx-community/silero-vad".to_string(),
        min_silence_duration_ms: 250,
        min_speech_duration_ms: 250,
    })?;
```

then for each new chunk that you get from a microphone you call `push` and see the events:
```rust
            match vad.push(&buffer) {
                Some(VadEvent::SpeechStarted) => {
                    println!("🟢 heard you, keep talking…");
                }
                Some(VadEvent::SpeechEnded) => {
                    println!(
                        "🔴 done — {:.2}s captured\n",
                        buffer.len() as f32 / sample_rate as f32
                    );
                    break;
                }
                None => {}
            }
```

I will do docs later.


nobodywho/core/src/vad/backend.rs
```diff
@@ -0,0 +1,148 @@
...
+const SILERO_SAMPLE_RATE: i64 = 16_000;
+
+pub(super) struct VadBackend {
+    session: Session,
+    sample_rate: u32,
+    /// LSTM hidden state carried between frames: shape (2, 1, 128) flattened.
+    model_state: Vec<f32>,
+    /// 16kHz f32 samples accumulated but not yet consumed as a full frame.
+    frame_buffer: Vec<f32>,
+    /// How many raw i16 samples of the caller's buffer we've already turned
```
duarteocarmo: perhaps a bit shorter-more concise comments.

nobodywho/core/src/voice_activity_detection/backend.rs
```diff
@@ -0,0 +1,148 @@
+use crate::errors::VadError;
+use crate::huggingface;
+use crate::onnx::{load_session, Device};
+use crate::stt::audio::{AudioResampler, DecodedAudio};
+use crate::vad::events::{DebounceConfig, Debouncer, VadEvent};
+use ort::session::Session;
+use ort::value::Tensor;
+
+/// Silero operates on exactly this many samples per frame at 16kHz.
```
duarteocarmo, line 14: Useless comment

duarteocarmo (changes requested):
Should we add tests?

marek-hradil:
@duarteocarmo I usually prefer to test it end-to-end from the integrations, as that feels what it means that it "works".

Since the integrations are not there yet (i will add them in the next steps), i did not add tests.

So I would do it that way, but what do you think?

duarteocarmo (approved):

AsbjornOlling (changes requested):
Can we update this to be consistent with the changes added in #623 ?

nobodywho/core/src/vad/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// `Vad` tracks the current turn internally). Returns `Some(VadEvent)`
+    /// if this call crossed a confirmed speech/silence boundary.
+    pub fn push(&mut self, chunk: &[i16]) -> Option<VadEvent> {
+        // push() only fails on ONNX inference errors, which would indicate
+        // a corrupt/incompatible downloaded model, not a caller error —
+        // surfacing that as a silently-swallowed None would hide a real
+        // bug, so unwrap here and let it surface loudly.
+        self.backend
+            .push(chunk)
+            .expect("Silero VAD inference failed — check the downloaded model is not corrupt")
```
AsbjornOlling: `backend.push` returns a `Result`, but this function swallows it into a panic.
Is it feasible to propagate the `Result` from here?

nobodywho/core/src/voice_activity_detection/backend.rs
```diff
@@ -0,0 +1,263 @@
...
+            Some(resampler) => resampler.push(&raw_f32)?,
+            None => raw_f32,
+        };
+        self.frames.extend(samples_16k);
+
+        let mut event = None;
+        while let Some(frame) = self.frames.next_frame() {
+            let prob = self.run_frame(&frame)?;
+            if let Some(e) = self.debouncer.step(prob) {
+                event = Some(e);
```
AsbjornOlling: If there are multiple events in one chunk, all except the last one are dropped.
This is *probably fine* in the streaming case with super small chunks, but there's no upper bound to chunk size, so if a user passes in a large chunk, they will get unexpected results.

As we spoke about IRL, a poth solution is to both:
a) add a new method that specifically handles long chunks with multiple events
b) `warn!` if multiple events are detected in this streaming-case method

nobodywho/core/src/vad/events.rs
```diff
@@ -0,0 +1,130 @@
...
+        self.state = State::Silence;
+    }
+
+    /// Feed one frame's speech probability, get back an edge event if this
+    /// frame crossed a confirmed speech/silence boundary.
+    pub fn step(&mut self, speech_prob: f32) -> Option<VadEvent> {
+        let min_speech_frames = (self.config.min_speech_duration_ms / FRAME_MS).max(1);
+        let min_silence_frames = (self.config.min_silence_duration_ms / FRAME_MS).max(1);
+        let is_speech = speech_prob >= self.config.threshold;
+        let is_silence = speech_prob < self.config.threshold - END_HYSTERESIS;
```
AsbjornOlling: Thresholds lower than `END_HYSTERESIS` (0.15) make it impossible to detect silence.
I think it's pretty likely that users will set random low numbers when experimenting.
We could emit a `warn!` if `config.threshold` is set to a too-low

AsbjornOlling (approved):

PR #623, opened 2026-07-10
New api API for onnx repo resolution.

AsbjornOlling:
## Motivation

Prior to this PR, I found that the onnx repo resolution was a little whack:

It would match against a fixed list of string literals to determine whether it should fetch from huggingface, and then fall back to local paths.
While we *do* only support a small subset of onnx repos, I think it would be nice to open up the option to also support downloading from forks and fine-tunes.
The model path was also inconsistent with what we do on the `Chat` and `Model` constructors. While it is fundamentally different, I think it would be good to be consistent with the urls, so it's very explicit to the user when something is downloading from hf versus when something is resolved locally. The logic of downloading from hf only if it matches a specific url in the source code is a bit hidden.
The "backend" terminology is a bit overloaded, since we typically use this to describe compute backends (e.g. metal vs vulkan).

## The changes

I renamed stuff called `backend` to be called `architecture` for tts architecture stuff.

Model architecture is automatically determined based on substring matching the repo path, but can be overriden by the `architecture` parameter. (Rather than matching specific urls).

I set it up to match on `hf` or `huggingface` url schemas to decide whether to download or resolve locally.

The diff for this PR is rather large because of regenerated files, but the actual changes to core is ~100 loc.

## The new API

 ```python
   import nobodywho

   # --- TTS ---

   # HuggingFace source — architecture auto-detected from "kokoro" in the path
   tts = nobodywho.Tts(source="hf://hexgrad/Kokoro-82M")
   wav = tts.synthesize("Hello from NobodyWho!")

   # HuggingFace source — architecture auto-detected from "supertonic" in the path
   tts = nobodywho.Tts(source="hf://Supertone/supertonic-3", voice="M1", language="en")
   wav = tts.synthesize("Hello from NobodyWho!")

   # Local directory — architecture must be specified (no "kokoro"/"supertonic" in path)
   tts = nobodywho.Tts(source="/path/to/custom-kokoro", architecture="kokoro")
   wav = tts.synthesize("Hello from NobodyWho!")

   # Override auto-detection with an explicit architecture
   tts = nobodywho.Tts(source="hf://my-org/my-custom-model", architecture="supertonic")

   # --- STT ---

   # HuggingFace source (Whisper is the only STT architecture, no architecture param needed)
   stt = nobodywho.STT(source="hf://onnx-community/whisper-base")
   text = stt.transcribe_file("recording.mp3").completed()

   # With options
   stt = nobodywho.STT(source="hf://onnx-community/whisper-base", language="en", quantization="q4")
 ```

## Alternatives

If we *really* do want to support downloading only one specific repo, rather than any repo of the same (very-specific) shape, we could consider simplifying the arguments to the Tts constructor to just be "kokoro" or "supertonic". Like `Tts("kokoro")`.

I think the middle-ground of *looking* like a HF path, but only supporting a specific HF path is misleading.


AsbjornOlling:
Looks like the full-ci dispatch is having issues. It's running here: https://github.com/nobodywho-ooo/nobodywho/actions/runs/29092225099

jona605a (approved):

PR #631, opened 2026-07-14
Fix /full-ci comment dispatch

duarteocarmo:
Apparently we can't trigger new workflows using GITHUB_TOKEN


jona605a (approved):
Consider if we want an error message if someone tries to do a CI on a PR from a fork.

PR #637, opened 2026-07-20
Feat/checkpointing

s194042:
Add checkpointing of recurrent state of hybrid models like Qwen3.5. The checkpointing in this PR is very simple and aims to improve the basic conversion flow. 

The main problem to deal with here is that think blocks are removed from the context when thinking is enabled. With the implementation on main this means a reread of the entire history on every ask. This PR addresses this by introducing a checkpoint after each prefill.


duarteocarmo:
Please rebase and (optionally) add an entry to `CHANGELOG.md` file. Thanks

PR #638, opened 2026-07-20
feat/Lark grammar and llguidance for tool calling

jona605a:
This PR looks at how sampling for tool calling is compiled, precomputed and activated in nobodywho. We are now using llguidance with lark grammars, and the code concerning sampling for tool calling has been refactored. 

## Changes.

**GBNF to Lark migration**. We previously used gbnf grammars for tool calling (which is still supported with `gbnf_to_lark`), and now grammar generation is changed to emit lark directly, across different model formats (Qwen3, Qwen3.5/3.6, Gemma4, FunctionGemma, LFM2, Ministral3). 

**llguidance**. We are now using llguidance for grammar sampling, which makes constrained generation much faster, especially for large vocabularies. 

**Pre-built samplers**. We now have two samplers / sampling states: free generation and grammar-constrained, with respective samplers `base_sampler` and `tool_sampler`. The base sampler is pre-built on creating a chat worker, and the tool sampler is also pre-built if tools are enabled, otherwise when tools are set. This way, we avoid the long (~400-1000 ms) startup time of compiling the tool sampler during a response. 

**Dynamic grammar activation**. During generation, sampling is performed using the `base_sampler` until a `<begin_tool_call>` token is generated, which switches in the `tool_sampler` instead for the remainder of that response. 

**"Slices" as a model-specific speedup**. Using constrained generation takes time to pre-compile the sampler and makes generation slower. In order to speed these up, "slices" can be given to the llguidance parser. I wrote a PR to llama-cpp-rs enabling us to select slices that match each model format. In short, when the parser checks for valid tokens and is in a position where almost all tokens are allowed (such as inside a string in a json), the validation step takes a long time, and you can give it a regex slice that allows the parser to accept all strings accepted by the regex at once, giving a speedup. These are also model-specific: since Qwen3 uses json for tool formatting, it needs json-tuned slices, whereas Gemma4 needs another. 
I have not yet looked into other parser optimizations than slices, but such configuration is exposed through a unified `llguidance_sampler`. 

**Improved error handling**. 

## Remaining todos:

**Cache the `tok_env` per model**. The llguidance sampler requires 1) A model's token environment, and 2) A grammar. The first only depends on the model and can thus be cached, so we don't need to re-compute it on every new chat worker. 

**Fix DRY / penalty history**. With dynamic tool-call activation, we are switching in a different sampler during generation, since lark / llguidance has no native trigger-work mechanism like GBNF's `grammar_lazy`. This means repetition penalties / DRY sampling loses the state. Fix could be to replay an amount of tokens to the sampler.

**Fix multi-token marker leak** (Claude-reported bug). If some `<begin_tool_call>` token is longer than a single token, `wrap_respond` (in inference.rs) leaks the first token(s). 

**Point to the official llama-cpp-rs once my PR has been accepted**. I have a small pending PR to llama-cpp-rs that fixes a bug where some models like Gemma4 wants to output another EOG token but the grammar masks it out. 

**Enable compatibility for MTP when tools are active**. Because MTP does lookahead and also since it samples many tokens at once, grammar activation in the middle becomes tricky. I'm considering making a single stateful sampler that switches between base ("free") sampling and tool (constrained) sampling. Deferred to a later PR.

### DIFF
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -38,6 +38,7 @@ use crate::tool_calling::{detect_tool_format, Tool, ToolCall, ToolFormat, ToolFo
 use ahash::AHasher;
 use indexmap::IndexMap;
 use llama_cpp_2::mtmd::MtmdBitmap;
+use llama_cpp_2::sampling::LlamaSampler;
 use llama_cpp_2::token::LlamaToken;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::cmp::min;
@@ -1380,14 +1381,15 @@ impl ChatContext {
 
 /// Builds the tool-call grammar sampler for an already-detected `tool_format`
 /// (detection happens once, in `new_chat_worker` — see the `tool_format`
-/// field doc). `Ok(None)` if `tools` is empty. `Err(DetectionFailed)` if
-/// tools are requested but no format was ever detected for this model.
+/// field doc), along with the begin-token sequence that triggers the switch to
+/// it. `Ok(None)` if `tools` is empty. `Err(DetectionFailed)` if tools are
+/// requested but no format was ever detected for this model.
 fn build_tool_sampler(
     model: &llama_cpp_2::model::LlamaModel,
     tools: &[Tool],
     sampler_config: &SamplerConfig,
     tool_format: Option<&ToolFormat>,
-) -> Result<Option<llama_cpp_2::sampling::LlamaSampler>, ToolCallingSetupError> {
+) -> Result<Option<(LlamaSampler, Vec<LlamaToken>)>, ToolCallingSetupError> {
     if tools.is_empty() {
         return Ok(None);
     }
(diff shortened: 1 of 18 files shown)


duarteocarmo:
Please rebase and (optionally) add an entry to `CHANGELOG.md` file. Thanks

nobodywho/core/src/chat.rs
```diff
@@ -1362,13 +1361,47 @@ impl ChatContext {
...
+}
+
 /// A chat session: owns an [`InferenceEngine`] plus all the conversational state
 /// (messages, tools, template, sampler config).
 struct Chat<'a> {
     engine: InferenceEngine<'a>,
     should_stop: Arc<AtomicBool>,
-    tool_grammar: Option<gbnf::GbnfGrammar>,
+    /// Is read from the model, used to control lark generation. None when
+    /// model detection fails.
```
s194042: These field description comments are unnecessary.

nobodywho/core/src/chat.rs
```diff
@@ -1393,48 +1426,62 @@ impl<'a> Chat<'a> {
...
+        };
+
+        let tool_sampler = build_tool_sampler(
+            &model.language_model,
+            &config.tools,
+            &sampler_config,
+            tool_format.as_ref(),
+        )?;
+
+        // Tool calling and MTP can't be combined yet, so force solo decode when
```
s194042: What? Why can it not?
Are we moving combining these two to another PR?

nobodywho/core/src/chat.rs
```diff
@@ -1615,9 +1661,16 @@ impl<'a> Chat<'a> {
         let mut full_response: String = String::with_capacity(4096);
         let mut tokens_written_until_now = TokenizerChunks::new();
 
-        // initialize sampler
-        // stateful samplers only live for one response
-        let mut sampler = sampler_config.to_stateful(self.engine.ctx.model)?;
+        // Reset stateful sub-sampler state (RNG, penalty/DRY history, grammar
```
s194042: This comment is also quite unnecessary. It is quite clear from the code here what is happening. Maybe look through the comments in and try to cut them down a bit.

nobodywho/core/src/tool_calling/qwen35_36.rs
```diff
@@ -31,29 +20,113 @@ impl ToolFormatHandler for Qwen35_36Handler {
...
 
+    // NOTE: intentionally no `slice_regexes` override. The hot position is the
+    // free-form `<parameter=…>` body, `body[suffix=/\n<\/parameter>\n/]:
+    // /(?s:.*)/`, whose `.*` compiles to a *lazy* lexeme. llguidance excludes
+    // lazy lexemes from slice subsumption (`RegexVec::subsume_possible` returns
+    // false whenever a lazy lexeme is in the state), so no slice regex can ever
+    // apply here — verified empirically (`slices_applied` stayed 0, trie walk
+    // unchanged). A useful slice would require expressing the body as a *greedy*
+    // bounded repeat (as FunctionGemma does), which we can't without banning the
+    // `<`, embedded/trailing newlines the suffix rule deliberately allows.
```
s194042, line 128: Well if now slice can ever apply, we should not apply the default slice, but rather supply no slices, correct? No use spending startup time building the default slice if it does not help.

s194042 (approved):

PR #641, opened 2026-07-21
feat: add support for pocket tts

### LINKED ISSUE #635: [FEATURE] Pocket-tts support
The provided two tts engines have their limitations and both do not allow to use samples to generate custom voice, pocket-tts runs on cpu and provides zero shot voice clone so it might be something to consider. 

Inclusion of kyutai pockettts engine similarly as kokoro and supertonic

As before godot can use udp to send the chat to external reader like in python but its less elegant than all in one nobodywho app, also there is a way to use a rust library to make a gdextension.

**Use case**
Same as with provided tts but with ability to voiceclone and maybe create a way to modify the voices ingame 

**Additional context**
the model in question : https://huggingface.co/kyutai/pocket-tts

duarteocarmo:
- adds pocket tts support 
- we need to support auth tokens since pocket has gated configuration 

Closes #635
NOB-141

### DIFF
--- a/nobodywho/core/src/tts/pocket.rs
+++ b/nobodywho/core/src/tts/pocket.rs
@@ -1,0 +1,644 @@
+use crate::errors::TtsError;
+use crate::tts::architecture::TtsArchitectureImpl;
+use crate::tts::TtsDevice;
+use ndarray::{Array, ArrayD, IxDyn};
+use ort::session::{Session, SessionInputValue, SessionOutputs};
+use ort::value::{DynTensor, DynValue, Tensor};
+use rand_distr::{Distribution, Normal};
+use safetensors::tensor::{Dtype, SafeTensors};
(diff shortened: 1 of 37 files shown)


nobodywho/core/src/tts/pocket.rs
```diff
@@ -0,0 +1,652 @@
...
+            temperature: DEFAULT_TEMPERATURE,
+            lsd_steps: 1,
+            huggingface_token: None,
+        }
+    }
+
+    fn huggingface_token(&self) -> Result<Option<String>, TtsError> {
+        if let Some(token) = &self.huggingface_token {
+            return Ok(Some(token.clone()));
+        }
```
s194042: Maybe HF Token auth belongs in huggingface.rs?
Might be out of scope of this PR.

s194042 (approved):
Needs a rebase, but other than that only one comment and it is non blocking. Looks good.

PR #644, opened 2026-07-26
sampler: run grammar before truncation samplers in GBNF presets

### LINKED ISSUE #421: [BUG] Structured output fails silently in Godot
**Describe the bug**
When I use `set_sampler_preset_json()` in my script to set the model to use structured output and run the program, it immediately quits once the scene is loaded, with no error message.
 
**To Reproduce**
1. Create a new scene
2. Use the simple chat template from the docs, put `set_sampler_preset_json()` right before the `ask()`
4. Run the scene
5. It immediately quits with no error

**Expected behavior**
The runtime should not quit, and the model should give structured output.

**Environment:**
 - OS: MacOS
 - Godot Version: 4.6.1
 - Nobodywho Version: 8.1.0
 - LLM Model: Tested with...
    - Phi-4-mini-instruct-Q4_K_M.gguf
    - Qwen_Qwen3-0.6B-Q4_K_M.gguf
    - Qwen_Qwen3-4B-Q4_K_M.gguf
    - gemma-3-4b-it-q4_0.gguf

**Additional context**
Here is my script:
```
extends NobodyWhoChat

func _ready():
    # self.response_updated.connect(_on_response_updated)
    self.response_finished.connect(_on_response_finished)

    start_worker()

    set_sampler_preset_json()
    ask("How are you?")

func _on_response_updated(token):
    # this will print every time a new token is generated
    print(token)

func _on_response_finished(response):
    # this will print when the entire response is finished
    print(response)
```
Here is the output:
```
2026-03-12T20:46:42.583325Z  INFO get_model: nobodywho::llm: Loading model use_gpu=true gpu_layers=4294967295 model_path="/Users/eric/Workspace/llm-test/models/gemma-3-4b-it-q4_0.gguf" use_gpu_if_available=true
2026-03-12T20:46:44.239808Z  INFO get_model:model_load: nobodywho::llm: Model loaded successfully model_path="/Users/eric/Workspace/llm-test/models/gemma-3-4b-it-q4_0.gguf" use_gpu_if_available=true path="/Users/eric/Workspace/llm-test/models/gemma-3-4b-it-q4_0.gguf"
2026-03-12T20:46:44.240963Z  INFO nobodywho::llm: Initializing worker
2026-03-12T20:46:44.474019Z  INFO nobodywho::chat: Worker processing: msg=SetSamplerConfig { sampler_config: SamplerConfig { steps: [TopK { top_k: 20 }, TopP { min_keep: 1, top_p: 0.95 }, Temperature { temperature: 0.6 }, Grammar { trigger_on: None, root: "root", grammar: "# this default gbnf grammar forces valid json output\nroot   ::= object\nvalue  ::= object | array | string | number | (\"true\" | \"false\" | \"null\") ws\n\nobject ::=\n\"{\" ws (\n            string \":\" ws value\n    (\",\" ws string \":\" ws value)*\n)? \"}\" ws\n\narray  ::=\n\"[\" ws (\n            value\n    (\",\" ws value)*\n)? \"]\" ws\n\nstring ::=\n\"\\\"\" (\n    [^\"\\\\\\x7F\\x00-\\x1F] |\n    \"\\\\\" ([\"\\\\bfnrt] | \"u\" [0-9a-fA-F]{4}) # escapes\n)* \"\\\"\" ws\n\nnumber ::= (\"-\"? ([0-9] | [1-9] [0-9]{0,15})) (\".\" [0-9]+)? ([eE] [-+]? [0-9] [1-9]{0,15})? ws\n\n# Optional space: by convention, applied in this grammar after literal chars when allowed\nws ::= | \" \" | \"\\n\" [ \\t]{0,20}" }], sample_step: Some(Dist), seed: 1234 } }
2026-03-12T20:46:44.476651Z  INFO nobodywho::chat: Worker processing: msg=Ask { text: "How are you?" }
2026-03-12T20:46:44.492472Z  INFO nobodywho::chat: Worker writing until done
--- Debugging process stopped ---
```

GuiBarradas:
Fixes #421

`SamplerPresets::json()` (and the deprecated `grammar()` preset) appended the Grammar step after the default TopK(20)/TopP/Temperature chain. When none of the candidates surviving top-k is grammar-valid, the grammar masks every token and llama.cpp throws a C++ exception. The exception unwinds across the FFI boundary, Rust cannot catch foreign exceptions, and the process aborts. In Godot the abort message never reaches the user, which is why the crash in #421 is silent.

This is also why the bug looks model-dependent: with Llama 3.2 the first constrained token (`{`) is inside the top 20, so the preset works. With thinking models like Qwen3-0.6B, which want to open with `<think>`, none of the top 20 candidates is valid for the JSON grammar and generation aborts. It matches the regression window too: the reports start at 8.1.0, and #396 is where `json()` became `default().shift(Grammar)`, moving the grammar behind the truncation steps.

The fix orders the grammar before the truncation samplers, so top-k only ever sees grammar-valid tokens.

Verified on Windows 11 (CPU build, cargo test):

- before: full generation with the json preset aborts with Qwen3-0.6B (`fatal runtime error: Rust cannot catch foreign exceptions, aborting`) and passes with Llama-3.2-3B
- after: passes with both, returning valid JSON
- the same A/B reproduces against the official Godot release DLLs (9.2.0 and 9.5.0, Godot 4.7.1 headless, Vulkan) using the 8-line repro from the issue

Tests added: json preset builds a sampler; full generation through the chat API with the json preset; and a model-independent ordering guard (a grammar requiring an unlikely literal plus `top_k(1)`, which forces the empty candidate set on any model if the ordering ever regresses).

Follow-up probably worth its own issue: hardening the FFI boundary so a grammar exception surfaces as an error instead of aborting the process. That lives in llama-cpp-rs rather than here.


jona605a (approved):
Nice PR

PR #649, opened 2026-07-28
Update CI to be gated.

s194042:
Our CI has long been way to thorough for must pushes/commits.
However we also want to keep the coverage across integrations, systems and so on.
To fix the first problem while keeping this condition mind, we introduce path gated jobs. The flow in simple terms is:
plan.yml: Detects which files have been changed and creates a set of booleans based on this.
->
main.yml: Uses the bools from plan.yml to decide which jobs to run and distributes the bools to the jobs that need to make further distinctions.
-> 
Any jobs that need to run are run

Once this PR has been approved I suggest we consider using a merge queue, so we each PR runs full-ci before merging. This will also require adding branch protection using required checks. This will depend on the job "required".


AsbjornOlling (approved):

AsbjornOlling:
Needs rebase after #648

krebso (approved):

PR #651, opened 2026-07-29
Fix CPU inference thread selection (make cpu go brrr)

duarteocarmo:
- Match `llama-cpp-python`'s CPU thread policy in the shared core: half the logical CPUs for token generation and all logical CPUs for prompt batches, including MTP contexts.
- Add an interleaved, CPU-only comparison benchmark using Gemma 4 E2B Q4_K_M on an Apple M3 Max; generation throughput excludes TTFT.

| Backend | Generation words/s | TTFT |
|---|---:|---:|
| NobodyWho | 44.17 ± 0.79 | 1.084 ± 0.548s |
| llama-cpp-python | 43.18 ± 0.71 | 1.242 ± 1.091s |


AsbjornOlling:
Chose to merge #650 instead.

PR #652, opened 2026-07-29
Add CPU inference performance benchmark

duarteocarmo:
- Add an interleaved CPU-only benchmark comparing NobodyWho with llama-cpp-python.
- Install NobodyWho from the `main` branch through uv inline metadata.
- Report generation throughput separately from time to first token across three seeds.

### DIFF
--- a/nobodywho/evals/performance/compare_llama_cpp.py
+++ b/nobodywho/evals/performance/compare_llama_cpp.py
@@ -1,0 +1,439 @@
+# /// script
+# requires-python = ">=3.12"
+# dependencies = [
+#   "nobodywho",
+#   "llama-cpp-python",
+# ]
+#
+# [tool.uv.sources]
+# nobodywho = { git = "https://github.com/nobodywho-ooo/nobodywho", branch = "main", subdirectory = "nobodywho/python" }
+# ///
+
+import argparse
+import random
+import time
+from dataclasses import dataclass
+from pathlib import Path
+from statistics import mean, stdev
+from typing import Literal
+
+from llama_cpp import Llama
+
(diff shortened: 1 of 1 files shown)


nobodywho/evals/performance/compare_llama_cpp.py
```diff
@@ -0,0 +1,339 @@
...
+    print("\n=== summary ===\n")
+    print(separator)
+    print(format_row(headers))
+    print(separator)
+    for row in rows:
+        print(format_row(row))
+    print(separator)
+
+
+def count_words_from_stream(text: str, pending: str) -> tuple[int, str]:
```
AsbjornOlling, line 253: Why count words? Shouldn't we count tokens instead?

nobodywho/evals/performance/compare_llama_cpp.py
```diff
@@ -0,0 +1,339 @@
...
+        new_words, pending_text = count_words_from_stream(token, pending_text)
+        completed_words += new_words
+        if completed_words >= OUTPUT_CUTOFF:
+            chat.stop_generation()
+            break
+
+    elapsed = time.perf_counter() - start
+    text = "".join(parts)
+    print(text, end="", flush=True)
+    result = make_result(
```
AsbjornOlling, line 299: Would be interesting to get average token of every token past the first one. The first token takes a lot longer

Currently it measures ttft and then tss is estimated to be (initialization_t+prefill_t+generation_t)/n_tokens.

duarteocarmo:
@AsbjornOlling  - added - as well as a couple of other improvements to the bench.

AsbjornOlling (approved):

PR #653, opened 2026-07-30
Release all bindings: new cpu core count, Pocket TTS support, automatic model selection, and gemma4 mtp support

AsbjornOlling:
Bumping versions, changelogs, docs and releasing


AsbjornOlling:
FWIW I dropped the React native STT change because it has already landed in the previous release, and it is *not* a breaking change, because the API was changed before the first version of the API released.

s194042 (approved):

AsbjornOlling:
Pushing tagged releases now

PR #654, opened 2026-07-30
Batch encoder and cross-encoder inputs

### LINKED ISSUE #625: [FEATURE] support llama_batch for embedding/reranking
**Problem**

CrossEncoder.rank() ([nobodywho/core/src/crossencoder.rs](https://github.com/nobodywho-ooo/nobodywho/blob/main/nobodywho/core/src/crossencoder.rs#L185)) loops over documents one at a time.
- Each iteration is a full context reset + a separate llama_decode() call
- On a Pixel 8a 40 documents reranking take 20s+ (that's too much)

Encoder.encode() has the same one-at-a-time shape for plain embeddings, no encodeBatch.

**Solution**

Expose llama.cpp's native multi-sequence batching for both Encoder and CrossEncoder — process multiple query/document pairs (or multiple texts, for Encoder) in a single llama_batch + one llama_decode()/ llama_encode() call, reading each result back out individually via llama_get_embeddings_seq(ctx, seq_id).

**Example**

llamadart does this for embedding: https://github.com/leehack/llamadart/blob/main/lib/src/backends/llama_cpp/llama_cpp_service.dart#L4459

It groups by however many tokens fit in n_batch and however many sequences llama_n_seq_max() allows, falling back to one-at-a-time only when a single input is too large or the model's pooling type doesn't support it.

**My use case**

Flutter app Android/iOS with semantic search with embedding and reranker.

duarteocarmo:
- Pack pooled encoder and cross-encoder inputs into llama.cpp multi-sequence batches within token and sequence limits.
- Preserve input order and fall back to single-sequence processing when pooling does not support batching.
- Expose `Encoder.encode_batch()` / `encodeBatch()` for all bindings
- Update docs
- Closes #625
- Closes NOB-144

### DIFF
--- a/nobodywho/core/src/llm.rs
+++ b/nobodywho/core/src/llm.rs
@@ -1,6 +1,10 @@
-use crate::errors::{InitWorkerError, LoadModelError, ReadError};
+#[cfg(test)]
+use crate::errors::ReadError;
+use crate::errors::{InitWorkerError, LoadModelError};
 use crate::huggingface::{download_gguf, parse_model_path};
-use crate::inference::{acquire_inference_lock, EngineContext, InferenceEngine};
+#[cfg(test)]
+use crate::inference::acquire_inference_lock;
+use crate::inference::{BatchCapacity, EngineContext, InferenceEngine};
 use crate::memory;
 use crate::model_selection;
 use crate::tokenizer::{ProjectionModel, Tokenizer};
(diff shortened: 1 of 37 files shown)


nobodywho/core/src/llm.rs
```diff
@@ -404,20 +430,18 @@ where
...
     }
 
-    /// Reset the KV cache and token count. Delegates to the inference engine.
-    pub fn reset_context(&mut self) -> &mut Self {
-        self.engine.reset_context();
-        self
-    }
-
     /// Tokenize `text` and read it into the context under the global inference lock.
+    #[cfg(test)]
```
krebso, line 444: **question**: what is the purpose of this?

krebso: is it just a method we use for testing?

duarteocarmo: Yeah, we only use it for unit testing on the `encoder.rs` - unrelated to this PR, but just a small cleanup.

krebso (approved):

Issue #647, opened 2026-07-27
[BUG] TTS in Godot fails with panic at `godot_core` with 'Godot async task failed'


pwab:
**Describe the bug**
When trying the new NobodyWhoTts node on slightly older AMD hardware with current Fedora OS I get an error at `var result: Dictionary = await tts.synthesize("Hello from NobodyWho!")` and my text is not synthesized.

Output:
```
Godot Engine v4.7.1.stable.official.a13da4feb - https://godotengine.org
OpenGL API 4.6 (Core Profile) Mesa 26.1.5 - Compatibility - Using Device: AMD - AMD Radeon RX 570 Series (radeonsi, polaris10, ACO, DRM 3.64, 7.1.5-200.fc44.x86_64)
```

Debugger:
```
E 0:00:05:389   godot_core::private::set_gdext_hook::{{closure}}: [panic godot/src/lib.rs:2140]
  there is no reactor running, must be called from the context of a Tokio 1.x runtime
  Context: Godot async task failed
  <C++ Source>  /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/godot-core-0.4.5/src/private.rs:343 @ godot_core::private::set_gdext_hook::{{closure}}()
```

**To Reproduce**
1. Go to the [official TTS docs](https://docs.nobodywho.ooo/godot/text-to-speech)
2. Try to recreate the **Quick start** sample

Or you can download my MRP: [nobodywho-tts.zip](https://github.com/user-attachments/files/30407542/nobodywho-tts.zip) (You need to download the nobodywho package and drop it into the `addons` folder as I did not wanted to upload a 4GB zip file.)

**Expected behavior**
- No errors
- Synthesized audio in user directory

**Environment:**
 - Hardware: AMD Ryzen 5 1600, 16,0 GiB RAM, AMD Radeon RX 570 8GB VRAM
 - OS: Fedora Linux 44
 - Godot Version: 4.7.1
 - Nobodywho Version: 9.5.0
 - LLM Model: Supertone/supertonic-3 (default, cached)

**Additional context**
None

madsmtm:
I think this was fixed by https://github.com/nobodywho-ooo/nobodywho/pull/718, and released in v11.0.0, feel free to reopen if that's not the case.

Issue #513, opened 2026-05-01
[FEATURE] [Godot] Calling tools from main thread


KNITLIS:
First of all thank you for your great tool! It made integrating llm into a game infinitely simpler than I anticipated before stumbling upon nobodywho

I propose this feature because of few issues what are very annoying to work around them and even then solutions leaves footguns in your project:

#### 1) Triggering redraw from a tool call
Nobodywho allows to integrate llm model into your game and quite reasonable scenario is running npc dialog with it. And when llm controls npc you want it's tools be able to interact with the world.
Lets take "door keeper" scenario as an example:

There is a door keeper in your game and player can ask them to open the door. They can do it with a tool call:
```gdscript
func _door_open_tool() -> String:
    if door.is_openable():
        door.texture = opened_door_texture
        return "door opened"
    return "door can't be opened"
```
This will throw an exception because you can't `queue_redraw()` from background thread. So you have to call `door.texture = opened_door_texture` deferred if you want to return conditional string from tool. Moreover in normal usecase tool handle is in npc script and door logic is in door script, so you just do `var result: String = door.open()` from tool function and then you need to design `door.open()` with llm tool call in mind, separating anything what would queue redraw to be called defered. 

#### 2) Awaiting signal from tool call
Sometimes you want llm powered npc to prompt player to do or "give" something to them with tool call:
```gdscript
signal item_given(item: ItemBase)

func _give_something_tool() -> String:
    GameUi.show_item_prompt()
    var item = await item_given
    return "Player gave you %s" % item.item_name
```
In that case your best choice is to create this monstrosity:
```gdscript
var tool_return_sema = Semaphore.new()
var tool_return_value: Variant

func _tool_wait() -> Variant:
	tool_return_sema.wait()
	return tool_return_value

func tool_return(answer: Variant):
	tool_return_value = answer
	tool_return_sema.post()

func _give_something_tool() -> String:
    GameUi.show_item_prompt.call_deferred()
    var item: ItemBase = await _tool_wait()
    return "Player gave you %s" % item.item_name
```
But it breaks type safety unless you're creating workaround for every answer type you can get.

---
So I propose moving execution of a tool call callable to main thread. Perhaps workaround from second example can become solution here:
```gdscript
# didn't coded enough in rust to give example in that language
var tool_return_sema = Semaphore.new()
var tool_return_string: String

func execute_tool(callable: Callable) -> String:
    _tool_wrapper.call_deferred()
[...]

Issue #203, opened 2025-08-06
[FEATURE] Allow async tool calls


emilnorsker:
Currently NobodyWho can not handle tool calls that are async.

I recently ran into this issue when implementing cross-encoders and having the ranking implemented as an async process. This is currently not possible due to a couple of issues.

Initially i tried to use the call_deferred method for Godot, as this function puts the execution at idle time in the end of a frame on the main thread. However it does not return anything, so to do that we need some kind of signal. That signal will have to be unique for the tool call, and we can't create it while in the worker thread (as it would mutate self and thus break our `Arc<dyn Fn>` requirement which needs 'static lifetime, but referencing self creates a closure tied to the method's lifetime instead of 'static).

The model will invoke the callable from a worker thread. Godot only allows certain operations (like using godot::task::spawn() and signal.to_future()) from the main thread, otherwise we get the error "godot_task() can only be used on the main thread". Additionally, the tool calling system stores closures as Arc<dyn Fn(serde_json::Value) -> String> which requires 'static lifetime, but any closure that captures references to self or scene nodes is tied to the method's lifetime, creating a fundamental Rust ownership conflict.

While using `call_deferred` successfully executes the async tool on the main thread, which does resolve the threading issues. However, retrieving the actual result is impossible within the current API constraints. Any solution that waits for async results requires storing state (channels or signals) which necessitates mutable references to objects or global state. Since tool closures must be 'static and cannot capture mutable references, there is no way to bridge the temporal gap between the immediate return required by the worker thread and the deferred execution that happens later on the main thread. This creates a fundamental architectural impossibility where async tool support cannot be implemented without changing the core tool interface.

**Describe the solution you'd like**
NobodyWho should be able to take an async function from Godot, use call_defered on it, have some kind of mechanism to get the actual FunctionState back and then wait for the GDScriptFunctionState's signal to complete, and receive the result from that. This likely requires fundamental changes to either the tool callback interface (moving away from Arc<dyn Fn(serde_json::Value) -> String>) or how tools are stored.

**Describe alternatives you've considered**
The only other solution is to throw an error when people try to do async. This is already on its way in pr #199. However, this significantly limits the usefulness of the tool calling system.

**Use case**
Having to wait for another process to complete before having data is a common usecase and should not be forced to be blocking.

examples:
- querying a db that takes some time.
- executing in-game actions that takes time to complete (walking to a spot, executing an animation etc)
- waiting for any kind of event to receive a message

**Additional context**
[...]

PR #658, opened 2026-07-31
Complete rewrite of godot bindings

AsbjornOlling:
A total rewrite of the godot bindings.

They had grown bloaty and ugly and totally illegible.

Should solve #513 and #203 and #647 at least

Still needs:
- [x] docs
- [ ] testing on android
- [x] VoiceActivityDetection
- [x] .complete function

## godot async kinda sucks

The core tricky thing with godot is how to handle async stuff. Godot does have some poorly-developed notion of coroutines, but it's very gdscript-specific, and it's (AFAIK) not possible to provide a gdscript-style coroutine via a gdextension (native code extensions for godot).

Async in godot is essential, because we can't block the main thread (that would mean dropping frames), and we can't rely on generating tokens faster than the target framerate.

The thing to use for async is *signals*. A godot signal is a thing that emits events. You can either hook a callback onto it, or you can await it (blocks the coroutine until an event was emitted). 

What we used to do was to have signals attached on the class itself, so you would do something like this:

```gdscript
chat.ask("hi")
await chat.response_completed
```

this works fine for things that are sort of singular: like generating tokens on a single sequence, but it gets tricky when it comes to stuff with multiple dispatch: e.g. kicking off a bunch of embeddings and waiting for their results. Tying the embeddings results back to the input text easily gets race-y and weird.

So now we instead create a new signal on each async function (wrapped by `NobodyWhoTask`) and return that signal.

A thing to watch out for is that these signals must be immediately awaited, or you may miss the signal. If a signal emits an event *before* you await it, you don't get the event you missed.
To make sure you always get the event, we intentionally emit the signal one frame late (with call_deferred).

So this will *always* catch the event x:

```gdscript
var x = await do_the_nw_thing()
```

and this will *also* always catch the event:

```gdscript
var x_signal = do_the_nw_thing()
var y = 2+2
var x = await x_signal
```

but you can't do something like this (it might miss the event, and then wait forever):

```gdscript
var x_signal = do_the_nw_thing()
await wait_one_frame()
var x = await x
```

## Tool call functions can now do more stuff.

Two new things to note here:

Tools may now be async. We detect when a function returns a GDScriptFunctionState and then await the completed signal on that before returning. Ta-da async tool support.

We always execute tools on the main thread. A bunch of scene APIs in godot can only be called from the main thread, and we used to call the tool from the LLM worker thread. Bad. Now we instead just pass up a small message to an async dispatcher task that's running on the main thread. The dispatcher task owns the tools and calls them when polled by the gdext async runtime.

## No more GUI configuration

We used to lean heavily into Godot's Node abstraction for making gui-configurable classes. That meant you could click a button and get a filepicker for your model file, and type the system prompt into a cute little text area.

A bunch of stuff wasn't neatly configurable from the gui. E.g. a sampler chain was just not of a shape that would neatly fit into godot's gui editors.

Keeping things in godot-owned fields (those can be gui-editable) also meant that they could be written to at any time. That means that keeping config for e.g. a chat node in sync with the actual worker thread got tricky. Things might change after starting the worker.
We had a pretty shitty solution of a `start_worker()` function that read all of the state and spawned a worker thread- but if you changed the settings afterwards the worker wouldn't necessarily be updated. Bad.

Now we just went code-only with the new godot bindings. Like in all our other bindings, you pass in the configuration at class construction time (with the .create call).

## a change to core

I added a thing to join worker threads when dropping them to the TTS and STT structs in core- this solves a reproducible segfault on teardown of the godot application.

## Lacking things added

A bunch of new stuff:

- .segment() for VAD
- bash() tool
- new introspection functions: `tokenize`, `mtp_acceptance_rate`
- streaming STT
- generalized chat template variables
- proper sampler builders
- proper tokenstream for ask (previous version was signals-based)
- clean teardown (join_on_drop in core)
- bumped gdext crate
- rewrote all of the docs. they now more closely match the other docs.


AsbjornOlling:
I suspect that full-ci is blocked because of a greedy-determinism-across-reset bug. Let's see if CI reproduces.

krobert:
Hey,  
Nice to see this split from monster libs.rs to small modules :D 

since you are refactoring the whole thing, i was wondering if you can expose more data from `read_gguf_model_info` and create a godot binding. this way i can create a model catalog for my downloaded models, showing more info before loading it, just by inspecting the file.
I also wanted to make `available_model_memory` public, so you can tell the user if they have memory to load that model. 

Question: 
Are you going to delete NobodyWhoDownloader and max_ctx func? those can be still usefull. 
image

Here is my suggestion:
in godot model.rs:
```
/// Inspect GGUF metadata without loading the model weights.
    #[func]
    fn inspect_gguf(path: GString) -> Variant {
        let path = resolve_godot_path(&path);
        let Some(info) = nobodywho::llm::inspect_gguf(Path::new(&path)) else {
            godot_error!("NobodyWhoModel.inspect_gguf: failed to inspect {}", path);
            return Variant::nil();
        };

        let mut dict = VarDictionary::new();
        dict.set("path", path.as_str());
        dict.set("file_size", info.file_size as i64);
        dict.set("tensor_count", info.tensor_count);
        dict.set("block_count", info.block_count as i64);
        dict.set("has_chat_template", info.has_chat_template);

        if let Some(value) = info.name {
            dict.set("name", value.as_str());
        }
        if let Some(value) = info.description {
            dict.set("description", value.as_str());
        }
        if let Some(value) = info.architecture {
            dict.set("architecture", value.as_str());
        }
        if let Some(value) = info.size_label {
            dict.set("size_label", value.as_str());
        }
        if let Some(value) = info.finetune {
            dict.set("finetune", value.as_str());
        }
        if let Some(value) = info.file_type {
            dict.set("file_type", value as i64);
        }
        if let Some(value) = info.quantization_version {
            dict.set("quantization_version", value as i64);
        }
        if let Some(value) = info.context_length {
            dict.set("context_length", value as i64);
        }
        if let Some(value) = info.embedding_length {
            dict.set("embedding_length", value as i64);
        }
        if let Some(value) = info.head_count {
            dict.set("head_count", value as i64);
        }
        if let Some(value) = info.head_count_kv {
            dict.set("head_count_kv", value as i64);
        }

        Variant::from(dict)
    }

    /// Returns memory currently available to NobodyWho for model loading.
    #[func]
    fn get_available_memory(use_gpu: bool) -> Variant {
        match nobodywho::llm::available_model_memory(use_gpu) {
            Ok((free_bytes, usable_bytes)) => {
                let mut dict = VarDictionary::new();
                dict.set("free_bytes", free_bytes as i64);
                dict.set("usable_bytes", usable_bytes as i64);
                Variant::from(dict)
            }
            Err(e) => {
                godot_error!("NobodyWhoModel.get_available_memory: {e}");
                Variant::nil()
            }
        }
    }
    #[func]
    fn is_generative_model(&self) -> bool {
        self.inner.is_generative_model()
    } # max_ctx etc.
```


llm.rs:
```
#[derive(Debug, Clone)]
pub struct GgufInspection {
    pub file_size: u64,
    pub tensor_count: i64,
    pub name: Option<String>,
    pub description: Option<String>,
    pub architecture: Option<String>,
    pub size_label: Option<String>,
    pub finetune: Option<String>,
    pub file_type: Option<u32>,
    pub quantization_version: Option<u32>,
    pub context_length: Option<u32>,
    pub embedding_length: Option<u32>,
    pub block_count: u32,
    pub head_count: Option<u32>,
    pub head_count_kv: Option<u32>,
    pub has_chat_template: bool,
}

pub fn inspect_gguf(path: &Path) -> Option<GgufInspection> {
    let info = memory::read_gguf_model_info(path)?;
    Some(GgufInspection {
        file_size: info.file_size,
        tensor_count: info.tensor_count,
        name: info.name,
        description: info.description,
        architecture: info.architecture,
        size_label: info.size_label,
        finetune: info.finetune,
        file_type: info.file_type,
        quantization_version: info.quantization_version,
        context_length: info.context_length,
        embedding_length: info.embedding_length,
        block_count: info.block_count,
        head_count: info.head_count,
        head_count_kv: info.head_count_kv,
        has_chat_template: info.has_chat_template,
    })
}

pub fn available_model_memory(
    use_gpu: bool,
) -> Result<(u64, u64), crate::errors::MemoryDetectionError> {
    let memory = memory::available_model_memory(use_gpu)?;
    Ok((memory.free_bytes, memory.usable_bytes))
}
```
memory.rs:
```
pub(crate) struct GgufModelInfo {
    pub n_layers: u32,
    pub file_size: u64,
    pub tensor_count: i64,
    pub name: Option<String>,
    pub description: Option<String>,
    pub architecture: Option<String>,
    pub size_label: Option<String>,
    pub finetune: Option<String>,
    pub file_type: Option<u32>,
    pub quantization_version: Option<u32>,
    pub context_length: Option<u32>,
    pub embedding_length: Option<u32>,
    pub block_count: u32,
    pub head_count: Option<u32>,
    pub head_count_kv: Option<u32>,
    pub has_chat_template: bool,
}

pub(crate) fn read_gguf_model_info(path: &Path) -> Option<GgufModelInfo> {
    let ctx = llama_cpp_2::gguf::GgufContext::from_file(path)?;
    let file_size = std::fs::metadata(path).ok()?.len();

    let find_u32 = |suffix: &str| {
        (0..ctx.n_kv())
            .find(|&i| ctx.key_at(i).is_some_and(|k| k.ends_with(suffix)))
            .map(|i| ctx.val_u32(i))
    };
    let find_str = |key: &str| {
        let idx = ctx.find_key(key);
        (idx >= 0)
            .then(|| ctx.val_str(idx))
            .flatten()
            .map(str::to_owned)
    };

    // llama.cpp counts offloadable layers as block_count + 1. The extra layer
    // is the output/embedding layer. If we only offload block_count layers,
    // layer 0 stays on CPU and every token requires a CPU<->GPU round-trip,
    // which can degrade performance by 3-30x depending on model size.
    let block_count = find_u32(".block_count")?;

    Some(GgufModelInfo {
        n_layers: block_count + 1,
        file_size,
        tensor_count: ctx.n_tensors(),
        name: find_str("general.name"),
        description: find_str("general.description"),
        architecture: find_str("general.architecture"),
        size_label: find_str("general.size_label"),
        finetune: find_str("general.finetune"),
        file_type: find_u32("general.file_type"),
        quantization_version: find_u32("general.quantization_version"),
        context_length: find_u32(".context_length"),
        embedding_length: find_u32(".embedding_length"),
        block_count,
        head_count: find_u32(".attention.head_count"),
        head_count_kv: find_u32(".attention.head_count_kv"),
        has_chat_template: ctx.find_key("tokenizer.chat_template") >= 0,
    })
}
```

nobodywho/godot/src/chat.rs
```diff
@@ -0,0 +1,766 @@
...
+        validate_config_keys(
+            config,
+            &[
+                "system_prompt",
+                "n_ctx",
+                "n_threads",
+                "use_gpu",
+                "sampler",
+                "template_variables",
+                "tools",
```
duarteocarmo, line 538: MTP option missing?

docs/docs-godot/faq.md
duarteocarmo, line 1: No more faq?

docs/docs-godot/index.md
duarteocarmo, line 1: Lets make sure this convention of naming organizing/pages is the same as the other bindings

docs/docs-godot/speech-to-text.md
```diff
@@ -1,80 +1,79 @@
...
 
-NobodyWho only supports Whisper models in **ONNX** format. `model_path` is a Hugging Face repo (`hf://owner/repo`) or a local directory containing such a model, e.g. `hf://onnx-community/whisper-base`. Browse the [Whisper ONNX models on Hugging Face](https://huggingface.co/models?library=onnx&search=whisper) to pick a size that fits your accuracy and speed needs.
+NobodyWho only supports Whisper models in **ONNX** format. The source is a Hugging Face repo
+(`hf://owner/repo`) or a local directory containing such a model, e.g.
+`hf://onnx-community/whisper-base`. Browse the
+[Whisper ONNX models on Hugging Face](https://huggingface.co/models?library=onnx&search=whisper)
+to pick a size that fits your accuracy and speed needs.
 
-You can also pick a `quantization` variant of the model to download and load. Lower-precision variants are smaller and faster, but can lose some transcription accuracy. Supported values are `default`, `fp16`, `int8`, `uint8`, `bnb4`, `q4`, `q4f16`, and `quantized`. Defaults to `default`.
+You can also pick a `quantization` variant of the model to download and load. Lower-precision
```
duarteocarmo, line 52: I think binding rejects this? - can you check?

nobodywho/godot/src/lib.rs
```diff
@@ -1,3779 +1,31 @@
...
+mod model;
+mod prompt;
+mod sampler;
+mod speech_to_text;
+mod task;
+mod text_to_speech;
+mod tools;
+mod vad;
+
+// --- Logging -----------------------------------------------------------------
```
duarteocarmo, line 19: Do we want these separator comments?

duarteocarmo (commented):
Did a first pass @AsbjornOlling  - some things to address already. 

Let's do a couple of passes I think easier.

docs/docs-godot/faq.md
AsbjornOlling: Nope. Making the docs match the other bindings.

https://github.com/nobodywho-ooo/nobodywho/pull/658#discussion_r4005236351

docs/docs-godot/speech-to-text.md
AsbjornOlling: Turns out that "fp16" and "q4f16" fail because of a known upstream bug in ort 2.0.0-rc.12
I dropped those from the docs. It's still an issue in the other bindings' docs. Made an issue on notion to fix properly in core.

All of the values listed are accepted by core, but literal `fp32` was missing from godot docs (only mentioned as `default`).

The rest of them work fine.

docs/docs-godot/index.md
AsbjornOlling: I don't understand. What is different?

duarteocarmo: order and naming of the pages. 

If they are all ok then you can ignore

nobodywho/godot/src/chat.rs
AsbjornOlling: Weird that I missed that. Good catch! It's implemented now.

AsbjornOlling:
@duarteocarmo review notes have been addressed, and have rebased on main to make mergable again

duarteocarmo (approved):

duarteocarmo:
@AsbjornOlling approved.
Just a slight merge conflict and you are good to go.

PR #667, opened 2026-08-06
fix: Context shift history measurement

duarteocarmo:
Context shifting deleted old messages from a local copy but repeatedly measured the unchanged original history


marek-hradil (approved):
Good work for spotting this.

PR #674, opened 2026-08-12
feat: Chat completion

jona605a:
opening a PR for now for doing CI.


jona605a:
Leaving as WIP - bindings not fully updated.

nobodywho/core/src/chat.rs
```diff
@@ -155,38 +101,158 @@ impl Message {
...
+        return Err(InvalidHistoryError::MisplacedSystemMessage { index: index + 1 });
+    }
+
+    validate_system_message_media(messages)
+}
+
+/// A leading system message is hoisted into the system prompt, which is stored
+/// as plain text — media in it would flatten to a marker with no bitmap behind
+/// it and fail deep inside tokenization.
+pub fn validate_system_message_media(messages: &[Message]) -> Result<(), InvalidHistoryError> {
```
madsmtm: Nit: I feel that merging this with `hoist_system_message` somehow would make more sense? That'd avoid having the logic split across two functions. "Parse, don't validate", as the idiom says.

nobodywho/core/src/chat.rs
```diff
@@ -155,38 +101,158 @@ impl Message {
...
+        _ => Ok(()),
+    }
+}
+
+/// Turns kept at the end of the history during a context shift; the first turn
+/// is always kept too.
+const PRESERVED_RECENT_TURNS: usize = 2;
+
+/// Indices of the user messages, i.e. the start of each conversational turn.
+/// Anything before the first index is a prefix that a context shift never touches.
```
madsmtm, line 205: Q: What does this sentence mean?

nobodywho/core/src/chat.rs
madsmtm: Relatedly, it might make sense to move the error handling in `validate_completion_messages` into `ChatMsg::SetChatHistory` / `ChatMsg::Complete`? And have their `output_tx` changed to contain the `Result`.

That way, I'd feel more confident that we're handling validation everywhere.

(In the case of `ChatMsg::Complete`, it'd probably end up as `Sender<Result<(), UnboundedSender<llm::WriteOutput>>>`... So I can see the argument that that's a bit ugly).

nobodywho/core/src/chat.rs
```diff
@@ -1518,6 +1722,7 @@ struct Chat<'a> {
     sampler: ChatSampler,
     sampler_config: SamplerConfig,
     messages: Vec<Message>,
+    system_prompt: Option<String>,
```
madsmtm, line 1735: Rambling:

How strongly do we need `Message` to be in a certain format? I kinda feel like the answer to that question guides a lot of design decisions of how we store the data.

For example, do we _require_ (as in, models will just plain not work) a user message before an assistant message? And that two user messages cannot after each other? If so, then perhaps we should parse it into a `messages: Vec<UserMessageAgentResponse>`?

My understanding is that this is not the case though, so I kinda feel like we shouldn't impose a specific format, _even for the system prompt_? Maybe just allow a system prompt anywhere (at least in the data model)? I'm guessing the model's jinja template will throw a warning if it can't handle it?

nobodywho/core/src/chat.rs
```diff
@@ -1651,92 +1855,52 @@ impl<'a> Chat<'a> {
...
+            // 1 <= n <= turn_starts.len() - 3, so the drain is never empty and
+            // never reaches the preserved turns.
+            let n = min(turns_to_delete, deletable);
+            messages.drain(turn_starts[1]..turn_starts[1 + n]);
+            turns_to_delete = turns_to_delete.saturating_mul(2);
         }
 
         self.messages = messages;
         Ok(())
     }
```
madsmtm, line 1906: Remark: I'm terrible at reading integer logic and loops like this, so I'll trust you that this function is correct ;)

nobodywho/core/src/chat.rs
```diff
@@ -23,108 +23,54 @@
...
-    }
-}
-
 #[derive(Deserialize, Serialize, Clone, Debug)]
 #[serde(tag = "role", rename_all = "lowercase")]
 pub enum Message {
     User {
         content: MessageContent,
-        #[serde(default, skip_serializing_if = "Vec::is_empty")]
-        assets: Vec<Asset>,
```
madsmtm: Remark: I think merging this into `MessageContent` makes a lot of sense.

nobodywho/core/src/content.rs
```diff
@@ -0,0 +1,411 @@
...
+/// would read back as something else is wrapped in `{"type": "raw", …}`.
+/// Tagging only when necessary leaves the wire format unchanged for payloads
+/// that do not collide, while guaranteeing content survives a round trip out of
+/// the history and back in.
+#[derive(Clone, Debug)]
+pub enum MessageContent {
+    Text(String),
+    Parts(Vec<ContentPart>),
+    Json(Value),
+}
```
madsmtm, line 117: I know it's pre-existing, but this does not really make sense to me? Why would we ever need anything other than `Vec<ContentPart>` here?

What is the "wire format" in this case? As I understand it, there is no wire format other than tokens, and these can be fully represented by `String`.

nobodywho/core/src/chat.rs
```diff
@@ -457,6 +523,59 @@ impl ChatHandle {
...
+
+    /// Answer a full message list, which replaces the chat history.
+    ///
+    /// The list is the whole conversation: it must be non-empty, end in a user or
+    /// tool message, and carry a system message only in front. A leading system
+    /// message becomes the chat's system prompt; a list without one keeps the
+    /// prompt the chat already had. The response is appended, and a following
+    /// `ask` continues from there.
+    ///
+    /// [`Options`] follows the same rule for the chat's other settings.
```
madsmtm, line 545: Remark: I disagree that `Options` should do this, I think it should be completely stateless from the user's perspective.

Or if not, then we shouldn't even allow setting options here (why would you when e.g. `chat.set_sampler(...)` does the same thing, and is clearer).

But that can be discussed in a follow-up, so I won't block on this.

docs/docs-python/vision.md
```diff
@@ -64,3 +64,46 @@ chat.ask(prompt).completed() # It's a dog and a penguin!
 
 That should be it! Beware though, that consuming images and audio can quickly drain the context,
 and larger context sizes may be needed for smooth usage.
+
+## Media in a message list
```
madsmtm, line 68: Q: This is only in the python docs?

madsmtm (approved):
I've only looked at core and briefly at the python docs, but that part looks good.

nobodywho/core/src/chat.rs
jona605a: A conversational "turn" just means "a user message and the following response". Counting it in turns makes the math of context shifting easier and makes more sense (at least to me) in general.

nobodywho/core/src/chat.rs
jona605a: I did both of these now. Regarding the error handling, now the other setters can error (as we talked about before). I'm deferring a general fix to another PR.

jona605a:
Some notes for future implementation or decisions I want to highlight:

Bindings-related API decisions:
- Python and Flutter take named arguments
- Kotlin, Swift and React Native take an `Options` object
- Godot uses a separate complete_with_options(), because gdext's `#[opt(default = …)]` accepts only `GodotImmutable` types and so can't default an object parameter; Godot also omits tools, since it registers those via `add_tool()`

The `Options` struct (argument to `complete()`) contains the sampler, template variables and tools for now. It is sticky: a `Some()` value overwrites the setting (state) in `Chat`, while a `None` uses the existing one. This is a matter of design choice and how stateless we want to be.

Similarly, the system prompt is a setting extracted from `messages[0]`, and `get_chat_history` does not return it. A message with `"role":"system"` errors now.

`Asset` is now gone and media is now stored in `ContentPart`.

Now `apply_options` calls both `set_sampler_config` and `set_tools`, both of which recompile the tool-calling grammar. Deferred fix.

PR #676, opened 2026-08-12
Android testing and other fixing

s194042:
Add android testing on devices via CI

> Each android integrations tests the currently released version and a source build
> They run once nightly and the source builds are required to pass before a release. 

Also fixed a linking issue for flutter and kotlin builds
removed required job as it is no longer used.


marek-hradil:
For the record. I will split this - it has multiple fixes combined and i want to be able to review each one of them.

PR #686, opened 2026-08-19
Hide generated files in diffs

madsmtm:
This should make it so that changes to the generated files are hidden by default in the diff view, which makes reviewing easier IMO.

As a side bonus, this should also make the "Languages" tab in the sidebar more descriptive. Currently it's:
Screenshot 2026-08-19 at 16 04 08

Where in reality, it's probably more Rusty.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+Cargo.nix linguist-generated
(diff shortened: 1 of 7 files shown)


AsbjornOlling (approved):

AsbjornOlling:
Love it!

madsmtm:
Not that I really think it'll be necessary, but to get a feel for it:

madsmtm:
Fair points clanker

PR #687, opened 2026-08-19
feat: Add `temp_ext`, `top_n_sigma` and `logit_bias` sampler steps

### LINKED ISSUE #572: [FEATURE] Expose remaining llama.cpp samplers (dynamic temperature, logit_bias, top_n_sigma)
## Summary

`SamplerBuilder` wraps most of llama.cpp's samplers, but a few useful ones aren't exposed in **any** binding. This proposes adding them.

## Missing samplers

Verified against the pinned `llama-cpp-2` fork (`marek-hradil/llama-cpp-rs`) — these `LlamaSampler` constructors exist upstream but are never called by core:

- **`temp_ext` — dynamic temperature (dynatemp)** — `temp_ext(t, delta, exponent)`. Entropy/range-based temperature; the most commonly requested of the three. We currently only expose flat `temp`.
- **`logit_bias` — boost / suppress / ban specific tokens** — `logit_bias(n_vocab, biases)`. Lets callers force or forbid particular tokens.
- **`top_n_sigma` — σ-based truncation** — `top_n_sigma(n)`. Niche but cheap to add.

All three are shift steps (they reshape the distribution), so they'd slot in alongside `top_k`, `min_p`, `temperature`, etc.

### Out of scope

- `grammar_lazy_patterns` — an internal building block for lazy-grammar triggering (tool calling), not a user-facing sampling knob.

## What exposing each entails

1. Add a `ShiftStep` variant + wiring in `build_step` in `nobodywho/core/src/sampler.rs`.
2. Thread it through each binding's `SamplerBuilder`:
   - Python — `nobodywho/python/src/lib.rs` (+ regenerate `.pyi`)
   - uniffi → Swift / React Native / Kotlin — `nobodywho/uniffi/src/lib.rs`
   - Flutter — `nobodywho/flutter/rust/src/lib.rs`
   - Godot — `nobodywho/godot/src/lib.rs`
3. Optionally add presets and docs (the sampling pages already list every step, so they'd just gain the new entries).

The recently-added `seed` step is a good template for the cross-binding wiring.

## Notes

- Distinct from #227 (combining/chaining samplers), which is already supported by `SamplerBuilder`. This issue is about samplers that aren't wrapped at all.
- `temp_ext` and `logit_bias` are the high-value items; `top_n_sigma` is a cheap extra.

madsmtm:
Add support for `temp_ext`, `top_n_sigma` and `logit_bias` sampler steps.

Fixes https://github.com/nobodywho-ooo/nobodywho/issues/572


nobodywho/core/src/sampler.rs
```diff
@@ -460,7 +476,37 @@ pub enum ShiftStep {
     Temperature {
         temperature: f32,
     },
+    /// Apply dynamic temperature scaling (a.k.a. entropy) described in the paper
+    /// <https://arxiv.org/abs/2309.02772>.
+    #[serde(rename = "temp_ext")]
+    TemperatureExtended {
```
madsmtm: Unsure about the naming here? We've renamed `temp` to `Temperature` above, so expanding the shorthand `ext` felt natural here as well.

It might make more sense to use `DynamicTemperature` though?

nobodywho/core/src/sampler.rs
```diff
@@ -460,7 +476,37 @@ pub enum ShiftStep {
...
+        n: f32,
+    },
+    /// Modify the likelihood of specific tokens.
+    LogitBias {
+        /// Mapping from token ID to its bias.
+        ///
+        /// The bias modifies the likelihood of the token being selected
+        /// (`>0.0` means higher probability of the token being selected).
+        /// Use [`f32::NEG_INFINITY`] to ban a token.
+        biases: HashMap<i32, f32>,
```
madsmtm, line 505: Pretty sure it's desirable to map it as a map, though ideally the token ID would be typed, instead of just an integer? `LlamaToken` isn't serializable though.

nobodywho/core/src/sampler.rs
duarteocarmo: `DynamicTemperature` makes more sense in my head. e.g., it says what it is. TempExt says little imo

duarteocarmo: (Would update bindings if we decide to go with it)

docs/docs-flutter/sampling.md
```diff
@@ -161,7 +161,10 @@ Shift steps — add as many as you want, applied in order:
 - `.typicalP(typP: 0.9, minKeep: 1)` — keep tokens whose "surprise" is close to average, dropping both the too-predictable and the too-random ([locally typical sampling](https://arxiv.org/abs/2202.00666))
 - `.xtc(xtcProbability: 0.5, xtcThreshold: 0.1, minKeep: 1)` — "exclude top choices": occasionally drop the top tokens for more variety
 - `.temperature(temperature: 0.8)` — below 1.0 = more focused, above 1.0 = more random
+- `.temperatureExtended(temperature: 0.8, delta: 0.5, exponent: 1.5)` — temperature in range [0.3; 1.3], with a variation of entropy^1.5
```
duarteocarmo: Perhaps add just 1/2 words on how it differs from temp for non technical?

duarteocarmo (commented):
I think pipes still need to pass!

madsmtm:
> I think pipes still need to pass!

Sorry, I should've waited for that. Also, do you think this needs `/full-ci`? What are the extra checks done by that?

duarteocarmo:
> > I think pipes still need to pass!
> 
> Sorry, I should've waited for that. Also, do you think this needs `/full-ci`? What are the extra checks done by that?

@madsmtm  yeah, that runs integration tests, generations, etc. It's a bit overkill and we can think how to make it faster. But I would run it!

nobodywho/core/src/sampler.rs
madsmtm: Changed it to `DynamicTemperature`, but actually, reading [the paper](https://arxiv.org/abs/2309.02772), they use the term "adaptive temperature", so maybe `AdaptiveTemperature` would be better?

On the other hand, llama.cpp / GGUF allows specifying the properties as `dynatemp_range` and `dynatemp_exponent`, so in that sense "dynamic" is superior.

Another option might also be to only expose `Temperature { t, delta, exponent }`, and when `delta == 0.0 && exponent == 1.0`, default to `LLamaSampler::temp`? And have the language bindings add default parameters when possible.

duarteocarmo: Would probably do what llama cpp does and go Dynamic in this case. But not a strong position!

docs/docs-flutter/sampling.md
madsmtm: Haha, I was being vague here because I didn't really understand how it worked myself. Looked into it a bit now, from what I gather it's useful for varying the temperature based on the model's confidence level (i.e. entropy).

The new description is:
> temperature in range [0.5; 1.1], scaled based on confidence level (`exponent` > 1.0 = when uncertain, higher temperature)

Lmk if that's still unclear?

marek-hradil (approved):

PR #608, opened 2026-07-06
feat: dynamicaly link GGML & llama.cpp

krebso:
## Description
Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce or add a test to the example game.

## Checklist:
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #691, opened 2026-08-21
feat: dynamic backends for android

krebso:
## Description
This PR implements dynamic backends feature from llama.cpp for android (kotlin, flutter, react-native). It is part of #608, but implementing (and testing) everything at once turned out to be a daunting task.

Previously, each Android ABI contained one statically linked GGML CPU backend
compiled for a conservative baseline (`armv8-a` or `x86-64`).

The new build uses upstream `GGML_CPU_ALL_VARIANTS` support. It packages the
shared GGML/llama.cpp runtime and several CPU backend. At startup, NobodyWho calls each variant’s `ggml_backend_score()` and loads
the best backend supported by the device.

This is CPU-only runtime selection. It does not select between CPU, Vulkan, and
OpenCL. Godot also keeps its
existing statically linked Android build.

**This gives us:**
- Multiple CPU implementations per ABI, using llama.cpp’s upstream backend
  module interface.
- Automatic selection of the best supported instruction set at runtime.
- GGML and llama.cpp shipped as shared libraries beside the binding.
- Matching `libc++_shared.so` packaged with the runtime.
- The same AAR works across supported old and new Android CPUs.


krebso:
pipeline passed, it failed because of some 503 when fetching vulkan

krebso:
tested locally by building the aar and having a small app on emulator, seems to be working:
```08-21 16:40:00.595 18548 18587 I nobodywho:    0.174365625s  INFO get_model{model_path="/data/user/0/ai.nobodywho.smoke.nobodywho_smoke/files/model.gguf" use_gpu_if_available=true mmproj_path=None draft_model_path=None}: nobodywho::llm: scored GGML CPU backend backend=/data/app/~~zOXPdP9HI-A8UOulub7eOg==/ai.nobodywho.smoke.nobodywho_smoke-V3mDZwYqURUNr5LmUfkaSQ==/base.apk!/lib/arm64-v8a/libggml-cpu-android_armv8.0_1.so score=1
08-21 16:40:00.596 18548 18587 I nobodywho:    0.175031416s  INFO get_model{model_path="/data/user/0/ai.nobodywho.smoke.nobodywho_smoke/files/model.gguf" use_gpu_if_available=true mmproj_path=None draft_model_path=None}: nobodywho::llm: scored GGML CPU backend backend=/data/app/~~zOXPdP9HI-A8UOulub7eOg==/ai.nobodywho.smoke.nobodywho_smoke-V3mDZwYqURUNr5LmUfkaSQ==/base.apk!/lib/arm64-v8a/libggml-cpu-android_armv8.2_1.so score=3
08-21 16:40:00.597 18548 18587 I nobodywho:    0.176406166s  INFO get_model{model_path="/data/user/0/ai.nobodywho.smoke.nobodywho_smoke/files/model.gguf" use_gpu_if_available=true mmproj_path=None draft_model_path=None}: nobodywho::llm: scored GGML CPU backend backend=/data/app/~~zOXPdP9HI-A8UOulub7eOg==/ai.nobodywho.smoke.nobodywho_smoke-V3mDZwYqURUNr5LmUfkaSQ==/base.apk!/lib/arm64-v8a/libggml-cpu-android_armv8.2_2.so score=7
08-21 16:40:00.598 18548 18587 I nobodywho:    0.177317208s  INFO get_model{model_path="/data/user/0/ai.nobodywho.smoke.nobodywho_smoke/files/model.gguf" use_gpu_if_available=true mmproj_path=None draft_model_path=None}: nobodywho::llm: scored GGML CPU backend backend=/data/app/~~zOXPdP9HI-A8UOulub7eOg==/ai.nobodywho.smoke.nobodywho_smoke-V3mDZwYqURUNr5LmUfkaSQ==/base.apk!/lib/arm64-v8a/libggml-cpu-android_armv8.6_1.so score=0
08-21 16:40:00.598 18548 18587 I nobodywho:    0.177672833s  INFO get_model{model_path="/data/user/0/ai.nobodywho.smoke.nobodywho_smoke/files/model.gguf" use_gpu_if_available=true mmproj_path=None draft_model_path=None}: nobodywho::llm: scored GGML CPU backend backend=/data/app/~~zOXPdP9HI-A8UOulub7eOg==/ai.nobodywho.smoke.nobodywho_smoke-V3mDZwYqURUNr5LmUfkaSQ==/base.apk!/lib/arm64-v8a/libggml-cpu-android_armv9.0_1.so score=0
08-21 16:40:00.599 18548 18587 I nobodywho:    0.177963875s  INFO get_model{model_path="/data/user/0/ai.nobodywho.smoke.nobodywho_smoke/files/model.gguf" use_gpu_if_available=true mmproj_path=None draft_model_path=None}: nobodywho::llm: scored GGML CPU backend backend=/data/app/~~zOXPdP9HI-A8UOulub7eOg==/ai.nobodywho.smoke.nobodywho_smoke-V3mDZwYqURUNr5LmUfkaSQ==/base.apk!/lib/arm64-v8a/libggml-cpu-android_armv9.2_1.so score=0
08-21 16:40:00.599 18548 18587 I nobodywho:    0.178430583s  INFO get_model{model_path="/data/user/0/ai.nobodywho.smoke.nobodywho_smoke/files/model.gguf" use_gpu_if_available=true mmproj_path=None draft_model_path=None}: nobodywho::llm: scored GGML CPU backend backend=/data/app/~~zOXPdP9HI-A8UOulub7eOg==/ai.nobodywho.smoke.nobodywho_smoke-V3mDZwYqURUNr5LmUfkaSQ==/base.apk!/lib/arm64-v8a/libggml-cpu-android_armv9.2_2.so score=0
08-21 16:40:00.600 18548 18587 I nobodywho:    0.179823458s  INFO get_model{model_path="/data/user/0/ai.nobodywho.smoke.nobodywho_smoke/files/model.gguf" use_gpu_if_available=true mmproj_path=None draft_model_path=None}: llama-cpp-2: loaded CPU backend from /data/app/~~zOXPdP9HI-A8UOulub7eOg==/ai.nobodywho.smoke.nobodywho_smoke-V3mDZwYqURUNr5LmUfkaSQ==/base.apk!/lib/arm64-v8a/libggml-cpu-android_armv8.2_2.so module="ggml::load_backend"
08-21 16:40:00.601 18548 18587 I nobodywho:    0.179959541s  INFO get_model{model_path="/data/user/0/ai.nobodywho.smoke.nobodywho_smoke/files/model.gguf" use_gpu_if_available=true mmproj_path=None draft_model_path=None}: nobodywho::llm: loaded GGML CPU backend backend=/data/app/~~zOXPdP9HI-A8UOulub7eOg==/ai.nobodywho.smoke.nobodywho_smoke-V3mDZwYqURUNr5LmUfkaSQ==/base.apk!/lib/arm64-v8a/libggml-cpu-android_armv8.2_2.so score=7```

File: nobodywho/core/cmake/llama-build-overrides.cmake [hunk 1]
```diff
@@ -0,0 +1,9 @@
+# CMAKE_PROJECT_INCLUDE is process-wide, so scope the overrides to the exact
+# llama.cpp Android build that needs loadable CPU backends.
+if(NOT ANDROID OR NOT GGML_BACKEND_DL OR NOT PROJECT_NAME STREQUAL "llama.cpp")
+  return()
+endif()
+
+set(LLAMA_OPENSSL OFF CACHE BOOL "" FORCE)
+set(GGML_CPU_ARM_ARCH "" CACHE STRING "" FORCE)
+set(CMAKE_PLATFORM_NO_VERSIONED_SONAME ON)
```
madsmtm: Could I get you to open a PR to https://github.com/utilityai/llama-cpp-rs adding the necessary support for this? We can do this for now, but long-term, I'd prefer to not have this workaround.

nobodywho/android/README.md
```diff
@@ -0,0 +1,115 @@
+# NobodyWho Android packaging
```
madsmtm, line 1: This file describes a lot of lower-level details and debugging info, which is nice, but could I also get you describe the high-level design of this + motivation? (or link me to somewhere on Linear/Notion where it's been discussed)

And maybe link to GGML's docs on it, or issues/PRs where it's been discussed in there?

In particular, I'd like to know:
- Why are we doing it like this vs. compiling different combinations of `ggml-cpu` with differently prefixed symbols, and linking them all together into one blob that selects the best backend automatically?
  - Dynamically dispatching to different SIMD implementations is a very common technique, so I'm surprised that it isn't applied here.
  - There's probably a good answer to this, but I'd like to see the arguments (and if there isn't a good answer, then I think we should fix it in `llama.cpp`).
- Is this only used for selecting between differently compiled CPU backends (with different SIMD instruction sets enabled), or is it also for figuring out Vulkan vs. OpenCL vs. CPU?
- Which `.so` files are we generating and how? How are they packaged? How are they chosen? Etc.

It'd also be nice to have the PR text detail how we were linking before vs. now?

nobodywho/core/src/llama_backend.rs
```diff
@@ -0,0 +1,115 @@
...
+
+#[cfg(all(feature = "android-dynamic-backends", target_os = "android"))]
+mod android {
+    use crate::errors::BackendInitError;
+    use std::ffi::{CStr, CString};
+    use std::path::Path;
+    use tracing::{debug, info};
+
+    pub(super) fn load_best_cpu_backend() -> Result<(), BackendInitError> {
+        let mut info: libc::Dl_info = unsafe { std::mem::zeroed() };
```
madsmtm: Nit: Avoid the `unsafe` by constructing the struct explicitly.
```suggestion
        let mut info = libc::Dl_info {
            dli_fname: ptr::null(),
            dli_fbase: ptr::null_mut(),
            dli_sname: ptr::null(),
            dli_saddr: ptr::null_mut(),
        };
```

nobodywho/core/src/llama_backend.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+#[cfg(all(feature = "android-dynamic-backends", target_os = "android"))]
+mod android {
+    use crate::errors::BackendInitError;
+    use std::ffi::{CStr, CString};
+    use std::path::Path;
+    use tracing::{debug, info};
+
+    pub(super) fn load_best_cpu_backend() -> Result<(), BackendInitError> {
```
madsmtm, line 29: Remark: Looks fairly sane, I like the idea of loading the backend relative to ourselves (this is sorta also what `@loader_path/...` does on macOS/iOS).

nobodywho/core/src/llama_backend.rs
```diff
@@ -0,0 +1,115 @@
...
+
+            if best_backend
+                .as_ref()
+                .is_none_or(|(_, _, best_score)| score > *best_score)
+            {
+                best_backend = Some((path, path_c, score));
+            }
+        }
+        let (backend_path, backend_path_c, score) =
+            best_backend.ok_or(BackendInitError::NoLoadableCpuBackend)?;
```
madsmtm, line 84: Remark: Is there some way that we can ensure that there's always a CPU backend available? Or would that cause symbol conflicts when trying to load the alternative CPU backends?

nobodywho/core/cmake/llama-build-overrides.cmake
madsmtm, line 9: Could I get you to open a PR to https://github.com/utilityai/llama-cpp-rs adding the necessary support for this? We can do this for now, but long-term, I'd prefer to not have this workaround.

nobodywho/android/build.gradle.kts
```diff
@@ -0,0 +1,186 @@
...
+        integration = "uniffi",
+        artifactId = "nobodywho-uniffi-android",
+        mainLibrary = "libnobodywho_uniffi.so",
+    )
+    else -> error("Unknown nobodywhoBinding '$binding'; expected flutter, react-native, or uniffi")
+}
+
+// Cargo's Android outputs in the layout build.yml uploads (see README):
+//   libnobodywho-<integration>-<target>-release.so
+//   nobodywho-runtime-<target>/*.so      common libs, CPU backends, libc++_shared.so
```
madsmtm, line 59: Huh, why is Cargo putting `libc++_shared.so` in here? Wouldn't that come from $ANDROID_HOME or whatever?

madsmtm (commented):
This one is hard! Good work though, I think that the overall approach is right (assuming that we can't use symbol prefixing instead, see first comment below).

I've left some initial comments, but we can also talk about it in person, probably Monday? (I'm likely going to be sick tomorrow as well)

nobodywho/android/build.gradle.kts
krebso: According to [the section shared runtimes](https://developer.android.com/ndk/guides/cpp-support), we need to package it ourselves.

nobodywho/core/cmake/llama-build-overrides.cmake
krebso: agreed

nobodywho/core/src/llama_backend.rs
krebso: there is always one such backend (that we would link statically with) without any specific instructions, and it gets score 1. i planned for this to handle the case where we mess up the packaging, and the resolution stops working. do you see how to do it more cleanly? i agree that it can get confused with the situation "none of the dynamically complied backends work".

nobodywho/core/src/llama_backend.rs
krebso: yup, that was my inspiration, also i am not sure how to do it any other way 😄

nobodywho/android/README.md
krebso: to answer your questions:
1. we use the `GGML_CPU_ALL_VARIANTS`, which tells llama.cpp to do it this way. i agree the latter is saner, but i do not think that it is compatible with the rest of the `load_backend` machinery from llama.cpp.
2. currently, you can only select from the cpu backends, but the premise is that we will be able to do also opencl and vulkan, we have been talking about this with Asbjørn.
3. good question, we could maybe document it somewhere, for arm, they are:
```
libggml-cpu-android_armv8.0_1.so
libggml-cpu-android_armv8.2_1.so
libggml-cpu-android_armv8.2_2.so
libggml-cpu-android_armv8.6_1.so
libggml-cpu-android_armv9.0_1.so
libggml-cpu-android_armv9.2_1.so
libggml-cpu-android_armv9.2_2.so
```
for x86 there are AVX, AVX2, AVX-512, VNNI, and AMX combinations IIRC. We always package all that llama.cpp produces with the feature flag turned on, and then we try all of them at runtime, selecting the highest scoring one.


I will also add the overview to the PR body.

madsmtm: > i do not think that it is compatible with the rest of the load_backend machinery from llama.cpp.

Yeah, true. Just wondering, because I pretty strongly believe that's the actually sane way of doing it. Idk., I don't want to block on it, but I also feel like we should at least have _tried_ to talk to the llama.cpp developers about it?

> currently, you can only select from the cpu backends, but the premise is that we will be able to do also opencl and vulkan, we have been talking about this with Asbjørn.

I don't think llama.cpp supports scoring non-CPU backends? Search for `GGML_BACKEND_DL_SCORE_IMPL` in llama.cpp, only the CPU backends seem to have that.

(And they don't actually "score", they just return a static value based on which CPU features are available on the system).

I somewhat recall talking about selecting other backends, but I don't think I understood why we don't want to choose Vulkan if it's supported? Is there ever a reason not to run on the GPU (with maybe OpenCL as the fallback)? Or is the plan that we implement our own scoring function for these backends?

> we could maybe document it somewhere

I think I'd like that, yes please!

nobodywho/core/src/llama_backend.rs
madsmtm: I guess if you want to be sure to catch situations where finding the backend stops working, then it doesn't make sense to have a fallback impl.

nobodywho/android/build.gradle.kts
madsmtm: I think I'm confused why it's in the Cargo directory?

PR #695, opened 2026-08-24
Improve logging on Android and iOS (and WASM)

### LINKED ISSUE #440: [FEATURE] Enable logs on Flutter

madsmtm:
Provide a common entry-point where loggers are initialized, and use platform-specific crates such as `android_logger` and `oslog` (and `console_log` for future WASM support) to initialize the logger.

The Python and Godot bindings do not use this, they integrate with the language's own logging framework instead.

Fixes https://github.com/nobodywho-ooo/nobodywho/issues/440. Replaces https://github.com/nobodywho-ooo/nobodywho/pull/517.

## How Has This Been Tested?

Tested in `flutter-start-example` on both iOS and Android, and in `NobodyWho-Chat` on iOS.

### DIFF
--- a/nobodywho/core/src/log.rs
+++ b/nobodywho/core/src/log.rs
@@ -1,0 +1,62 @@
+pub use log::LevelFilter;
+
+/// Configure a sensible global [logger][`log`] for the current target.
+///
+/// This also captures [`tracing`] events as long as either:
+/// - No tracing subscriber is registered.
+/// - The `tracing/log-always` feature is enabled.
+///
+/// NOTE: Ideally, we'd probably want to make this use `tracing` instead, and
+/// set up a `tracing_log::LogTracer` for catching `log` events. But for now,
+/// this will suffice.
+///
+/// This should be safe to call in a constructor.
+///
+/// # Panics
+///
+/// Panics if called more than once.
+pub fn init(level: LevelFilter) {
+    send_llamacpp_logs_to_tracing();
+
+    // Android logs to `adb logcat`.
+    #[cfg(target_os = "android")]
+    {
+        let config = android_logger::Config::default()
+            .with_max_level(level)
+            .with_tag("nobodywho");
+        android_logger::init_once(config)
+    }
+
+    // iOS/tvOS/watchOS/visionOS logs to oslog (/usr/bin/log)
+    #[cfg(all(target_vendor = "apple", not(target_os = "macos")))]
+    {
+        oslog::OsLogger::new("nobodywho")
+            .level_filter(level)
+            .init()
+            .expect("log initialization must only happen once")
+    }
+
+    // Web platforms log to the console.
+    #[cfg(target_family = "wasm")]
+    {
+        console_log::init_with_level(level).expect("log initialization must only happen once")
+    }
+
+    // All other platforms log to stderr.
+    #[cfg(not(any(
+        target_os = "android",
+        all(target_vendor = "apple", not(target_os = "macos")),
+        target_family = "wasm"
+    )))]
+    {
+        env_logger::Builder::new()
+            .filter_level(level)
+            .parse_default_env()
+            .init()
+    }
+}
+
+/// Forward `llama.cpp` logs into tracing.
+pub fn send_llamacpp_logs_to_tracing() {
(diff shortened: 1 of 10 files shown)


nobodywho/flutter/rust/src/lib.rs
```diff
@@ -1679,24 +1679,20 @@ impl SamplerPresets {
 
 #[flutter_rust_bridge::frb(init)]
 pub fn init_app() {
-    // send llamacpp logs into tracing
-    nobodywho::send_llamacpp_logs_to_tracing();
-
-    // send logs to the appropriate places for android, ios and wasm
-    flutter_rust_bridge::setup_default_user_utils();
```
madsmtm, line 1686: We don't want to call `setup_default_user_utils`, [the implementation](https://github.com/fzyzcjy/flutter_rust_bridge/blob/8631db91a90c699439c9ce2a1fb4164ae8661da9/frb_rust/src/misc/user_utils.rs#L35-L46) is a bit lacking, and is missing a default `stderr` logger for non-Android/iOS/WASM, so we'd rather handle this ourselves.

We might want to call [`flutter_rust_bridge::enable_frb_rust_to_dart_logging!()`](https://cjycode.com/flutter_rust_bridge/guides/how-to/logging#case-2-builtin-rust-to-dart-logging) in the future though, but it should probably be configurable in user code.

nobodywho/flutter/rust/src/lib.rs
```diff
@@ -1679,24 +1679,20 @@ impl SamplerPresets {
...
+            nobodywho::log::LevelFilter::Info
+        };
+        nobodywho::log::init(level);
 
-    tracing_subscriber::fmt()
-        .with_max_level(log_level)
-        .with_timer(tracing_subscriber::fmt::time::uptime())
-        .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE)
-        .try_init()
-        .ok();
```
madsmtm, line 1699: _Removing_ this is actually the fix; this allows `tracing` logs to be routed to `android_logger` instead.

nobodywho/core/src/log.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    // iOS/tvOS/watchOS/visionOS logs to oslog (/usr/bin/log)
+    #[cfg(all(target_vendor = "apple", not(target_os = "macos")))]
+    {
+        oslog::OsLogger::new("nobodywho")
+            .level_filter(level)
+            .init()
+            .expect("log initialization must only happen once")
+    }
```
madsmtm, line 37: Logs now show up like this in Xcode:

Image

Previously, it was quite garbled (because we were outputting ANSI color codes to stderr, which Xcode doesn't support).

nobodywho/core/src/log.rs
```diff
@@ -0,0 +1,62 @@
...
+    send_llamacpp_logs_to_tracing();
+
+    // Android logs to `adb logcat`.
+    #[cfg(target_os = "android")]
+    {
+        let config = android_logger::Config::default()
+            .with_max_level(level)
+            .with_tag("nobodywho");
+        android_logger::init_once(config)
+    }
```
madsmtm, line 28: Logs now show up like this in `flutter run`:

Image

nobodywho/flutter/rust/src/lib.rs
madsmtm: I tend towards thinking that we want to run `enable_frb_rust_to_dart_logging!(setup_dart_logging_output = false)`, and tell users to enable Dart logging in their app, though that means we'll have to update `flutter_rust_bridge`.

I'm also running into https://github.com/fzyzcjy/flutter_rust_bridge/issues/3402.

madsmtm: I went with this approach instead, and filed https://github.com/nobodywho-ooo/nobodywho/pull/698.

madsmtm:
I've thought about it, and decided to split this into https://github.com/nobodywho-ooo/nobodywho/pull/698 and https://github.com/nobodywho-ooo/nobodywho/pull/699, it makes sense to integrate with the language's logging features where we can (so the Python and Godot bindings are already doing roughly the right thing).

PR #696, opened 2026-08-26
Add NobodyWho agent skill and LLM docs

duarteocarmo:
- Add an installable NobodyWho agent skill and link it from the documentation.
- Include Kotlin and latest release docs in the LLM outputs, with normalized Markdown and release versions.


docs/plugins/llms-txt/index.js
```diff
@@ -48,13 +62,91 @@ function getTitle(filePath, content, frontmatter) {
   return path.basename(filePath, path.extname(filePath));
 }
 
+function absoluteDocsUrl({pageUrl, target}) {
+  const url = new URL(target, `${SITE_URL}${pageUrl}`);
+  url.pathname = url.pathname.replace(/\.mdx?$/, '');
+  return url.toString();
+}
+
+function normalizeMarkdown({content, pageUrl, title}) {
```
madsmtm, line 71: Remark: This _feels_ brittle, but I guess the specifics don't matter too much, the LLM will likely understand it regardless.

docs/README.md
```diff
@@ -111,7 +111,7 @@ The site deploys to Cloudflare Pages via `.github/workflows/docs.yml`. It trigge
 ## LLM-Friendly Output
 
 The build generates two files:
-- `build/llms.txt` — page index with links
-- `build/llms-full.txt` — full concatenated documentation
+- `build/llms.txt` contains an index with links to the shared docs and each binding's latest release docs.
+- `build/llms-full.txt` contains the shared docs and the full contents of each binding's latest release docs, grouped by binding and page. Each binding section states its release version.
```
madsmtm, line 115: Remark: Pre-existing, but I kinda feel like `llms-full.txt` isn't that useful, instead you'd rather want just the language-specific bindings? Would probably fill the context with more relevant stuff.

madsmtm (approved):
I think it's a good idea, I don't have much experience with agent skills though, so there's probably things here that I wouldn't know whether they're right or wrong.

docs/README.md
duarteocarmo: That's a good point! Not sure how to address it now - so will merge.

PR #697, opened 2026-08-28
Update `flutter_rust_bridge` to `2.13.0`

### LINKED ISSUE #440: [FEATURE] Enable logs on Flutter

madsmtm:
Needed for `enable_frb_rust_to_dart_logging!`, which would probably be the nicer way to fix https://github.com/nobodywho-ooo/nobodywho/issues/440.


s194042 (approved):
There is a conflict in the changelog, but otherwise looks good.

PR #698, opened 2026-08-28
Forward Flutter logs to Dart `logging` package

### LINKED ISSUE #440: [FEATURE] Enable logs on Flutter

madsmtm:
Integrate with Dart's logging mechanism, which allows surfacing logs in `flutter run`.

Depends on https://github.com/nobodywho-ooo/nobodywho/pull/697. Fixes https://github.com/nobodywho-ooo/nobodywho/issues/440. Replaces https://github.com/nobodywho-ooo/nobodywho/pull/517 and https://github.com/nobodywho-ooo/nobodywho/pull/695.

~Waiting on fix for https://github.com/fzyzcjy/flutter_rust_bridge/issues/3402.~ I've decided to implement a workaround in the meantime with `#[cfg(broken_logging)]`, we can remove that once `flutter_rust_bridge` v2.14.0 is released.

### DIFF
--- a/nobodywho/flutter/rust/src/lib.rs
+++ b/nobodywho/flutter/rust/src/lib.rs
@@ -1784,26 +1784,37 @@ impl SamplerPresets {
     }
 }
 
+// https://cjycode.com/flutter_rust_bridge/guides/how-to/logging
+// Logs path: `tracing` -> `log` -> Dart logging.
+flutter_rust_bridge::enable_frb_rust_to_dart_logging!(
+    max_level = if cfg!(debug_assertions) {
+        log::LevelFilter::Debug
+    } else {
+        log::LevelFilter::Info
+    },
+    // Not registering a default output is probably the better choice for
+    // a library like ours. This allows the user to customize the logging
+    // in their application.
+    setup_dart_logging_output = false
+);
+
+// FIXME(madsmtm): Remove the need for this?
+// https://github.com/fzyzcjy/flutter_rust_bridge/issues/3402
+impl frb_generated::SseEncode for FrbLogRecord {
+    fn sse_encode(self, _serializer: &mut flutter_rust_bridge::for_generated::SseSerializer) {
+        unimplemented!("serializing FrbLogRecord should not be needed")
+    }
+}
+
 #[flutter_rust_bridge::frb(init)]
 pub fn init_app() {
     // send llamacpp logs into tracing
     nobodywho::send_llamacpp_logs_to_tracing();
 
-    // send logs to the appropriate places for android, ios and wasm
-    flutter_rust_bridge::setup_default_user_utils();
+    // NOTE: Do not set up a `tracing-subscriber`, that will conflict with
+    // `enable_frb_rust_to_dart_logging!` above!
 
-    let log_level = if cfg!(debug_assertions) {
-        tracing::Level::DEBUG
-    } else {
-        tracing::Level::INFO
-    };
-
-    tracing_subscriber::fmt()
-        .with_max_level(log_level)
-        .with_timer(tracing_subscriber::fmt::time::uptime())
-        .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE)
-        .try_init()
-        .ok();
+    flutter_rust_bridge::setup_backtrace();
 }
 
 #[cfg(test)]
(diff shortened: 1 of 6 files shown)


nobodywho/flutter/rust/src/lib.rs
```diff
@@ -1793,24 +1793,32 @@ impl SamplerPresets {
...
-    };
+    // NOTE: Do not set up a `tracing-subscriber`, that will conflict with
+    // `enable_frb_rust_to_dart_logging!`.
 
-    tracing_subscriber::fmt()
-        .with_max_level(log_level)
-        .with_timer(tracing_subscriber::fmt::time::uptime())
-        .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE)
-        .try_init()
-        .ok();
```
madsmtm, line 1843: _Removing_ the tracing subscriber is actually part of the fix; this allows `tracing` logs to be routed via `log` instead because of the `tracing/log` feature being enabled.

docs/docs-flutter/logging-and-troubleshooting.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Logging and Troubleshooting
+sidebar_position: 8
```
jona605a, line 3: From my limited testing with running the docs with npm, it seems that `sidebar_position` does nothing, and instead this page needs to be added to `docs/sidebars/flutter.ts` in order to show up.

nobodywho/flutter/rust/src/lib.rs
```diff
@@ -1798,26 +1798,123 @@ impl SamplerPresets {
...
+flutter_rust_bridge::enable_frb_rust_to_dart_logging!(
+    max_level = if cfg!(debug_assertions) {
+        log::LevelFilter::Debug
+    } else {
+        log::LevelFilter::Info
+    },
+    // Not registering a default output is probably the better choice for
+    // a library like ours. This allows the user to customize the logging
+    // in their application.
+    setup_dart_logging_output = false
```
jona605a, line 1812: I'm unsure what the correct approach is here, so I'll trust your instinct

nobodywho/flutter/nobodywho/pubspec.yaml
```diff
@@ -35,6 +35,7 @@ dev_dependencies:
   test: ^1.25.6
   freezed: ^3.2.3
   build_runner: ^2.4.13
+  logging: 1.3.0
```
jona605a: (from ai-review:) this is the only line that doesn't use `^`. Should it?

nobodywho/flutter/nobodywho/test/doctest_generated_test.dart
```diff
@@ -334,6 +335,10 @@ void main() {
       print(msg); // Yes, indeed, water is wet!
     });
 
+    test('logging-and-troubleshooting.md:13', () async {
+      await _doctest_25();
```
jona605a, line 339: just noticed that this generates doctest code from the `logging-and-troubleshotting.md` file - wtf. Not sure how this affects logging in the rest of this generated doctest file or if that's a problem. Just a weird find. Feel free to ignore.

jona605a (approved):
I don't have a lot of experience / opinions on logging, so I trust your decisions here.

nobodywho/flutter/nobodywho/pubspec.yaml
madsmtm: Prolly, dunno flutter dep specifications that well

nobodywho/flutter/nobodywho/test/doctest_generated_test.dart
madsmtm: Yeah, maybe we should just disable doctests for that file / that example?

docs/docs-flutter/logging-and-troubleshooting.md
madsmtm: Ah, yeah, I didn't actually build the docs locally and check

nobodywho/flutter/rust/src/lib.rs
madsmtm: It is kinda annoying that we don't log by default, but the alternative would be to set this to `true`, which from my understanding there's no way to remove again, and that would probably be more annoying.

nobodywho/flutter/nobodywho/test/doctest_generated_test.dart
jona605a: Yeah, probably fine

PR #699, opened 2026-08-28
Improve logging in UniFFI crates

madsmtm:
Keep using `android_logger` on Android, use `oslog` on iOS, prepare for WASM support by using `console_log`, and default to logging to stderr everywhere else.

Found while doing https://github.com/nobodywho-ooo/nobodywho/pull/698.

### DIFF
--- a/nobodywho/uniffi/src/lib.rs
+++ b/nobodywho/uniffi/src/lib.rs
@@ -1,29 +1,73 @@
 use std::collections::HashMap;
 use std::path::PathBuf;
-use std::sync::Arc;
+use std::sync::{Arc, Once};
+
+use tracing::{debug, error, info};
 
 uniffi::setup_scaffolding!("nobodywho");
 
-/// Initialize logging so tracing output goes to Android logcat.
-#[cfg(target_os = "android")]
-fn init_logging() {
-    use std::sync::Once;
+pub fn init_logging() {
+    let level = if cfg!(debug_assertions) {
+        log::LevelFilter::Debug
+    } else {
+        log::LevelFilter::Info
+    };
+
     static INIT: Once = Once::new();
+
+    // Configure a sensible global logger.
+    //
+    // This also captures `tracing` events because the `tracing/log` feature
+    // is enabled, and because we don't register a tracing subscriber.
+    //
+    // NOTE: Ideally, we'd probably want to make this use `tracing` instead,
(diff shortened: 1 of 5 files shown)


nobodywho/uniffi/src/lib.rs
```diff
@@ -1,30 +1,74 @@
...
+            android_logger::init_once(config)
+        }
+
+        // iOS/tvOS/watchOS/visionOS logs to oslog (/usr/bin/log)
+        #[cfg(all(target_vendor = "apple", not(target_os = "macos")))]
+        {
+            oslog::OsLogger::new("nobodywho")
+                .level_filter(level)
+                .init()
+                .expect("log initialization must only happen once")
```
madsmtm, line 45: Logs now show up like this in Xcode:

Image

Previously, it was quite garbled (because we were outputting ANSI color codes to stderr, which Xcode doesn't support).

nobodywho/uniffi/src/lib.rs
```diff
@@ -1,30 +1,74 @@
...
-        log::info!("NobodyWho logging initialized");
+        nobodywho::send_llamacpp_logs_to_tracing();
+
+        // Android logs to `adb logcat`.
+        #[cfg(target_os = "android")]
+        {
+            let config = android_logger::Config::default()
+                .with_max_level(level)
+                .with_tag("nobodywho");
+            android_logger::init_once(config)
```
madsmtm, line 36: Logs show up roughly like this in `adb logcat`:

Image

nobodywho/uniffi/src/lib.rs
```diff
@@ -1,30 +1,74 @@
...
+    // we'd rather depend on more well-tested crates.
     INIT.call_once(|| {
-        // Set up android_logger so log crate macros go to logcat
-        android_logger::init_once(
-            android_logger::Config::default()
-                .with_max_level(log::LevelFilter::Debug)
-                .with_tag("nobodywho"),
-        );
-        // Bridge tracing → log crate → android_logger → logcat
-        tracing_log::LogTracer::init().ok();
```
madsmtm, line 20: The comment here is wrong, that's not what `LogTracer` does; it bridges `log` -> `tracing`!

The right thing to do is to enable the `tracing/log` Cargo feature (which bridges `tracing` -> `log`).

jona605a (approved):
Nice find! Looks like a good solution

PR #701, opened 2026-08-28
fix: setter errors no longer kill the chat worker

jona605a:
A setter the worker rejected propagated out of `process_worker_msg`, which both worker loops treated as fatal. One bad argument ended the thread: the caller saw only "worker terminated", the cause went to the log, and every later call on that chat failed the same way. PR #674 fixed this for `set_chat_history`; this covers the rest and removes the crash path.

- every setter reply channel carries `Result<(), SetterError>`
- `process_worker_msg` is infallible; no arm can end the worker
- `SetterError` gains typed variants and a `Diagnostic` derive
- godot emits `worker_failed`; the sampler presets used to discard the error
- flutter setters return `String`, dropping the opaque Dart `SetterError`
- test: rejected sampler surfaces, old config intact, `ask()` still answers


jona605a:
Full ci passed after one test was re-run (download error)
Changelog issue fixed.

nobodywho/core/src/chat.rs
```diff
@@ -571,15 +569,23 @@ impl ChatHandle {
...
+    /// Send a setter message and block until the worker answers. The outer error
+    /// is the worker being gone, the inner one is the worker rejecting the value.
+    fn set_and_wait_blocking<F>(
+        &self,
+        make_msg: F,
+        label: &str,
+    ) -> Result<(), crate::errors::SetterError>
     where
-        F: FnOnce(tokio::sync::mpsc::Sender<()>) -> ChatMsg,
+        F: FnOnce(SetterReply) -> ChatMsg,
```
madsmtm, line 580: Remark: I initially though I'd prefer that this was generic over the result type instead:

Something like:

```suggestion
    fn set_and_wait_blocking<R>(
        &self,
        make_msg: impl FnOnce(tokio::sync::mpsc::Sender<R>) -> ChatMsg,
    ) -> Option<R>
```

That way, we don't have to shove everything into `SetterReply`.

But re-reading, I see that we're also handling the error from `output_tx.blocking_recv()`, and I can see the point of doing that centrally.

nobodywho/core/src/errors.rs
```diff
@@ -469,16 +469,29 @@ pub enum WorkerError {
     GILPoison, // this is actually a std::sync::PoisonError<std::sync::MutexGuard<'static, ()>>, but that doesn't implement Send, so we do this
 }
 
-#[derive(Debug, thiserror::Error)]
+/// A setter's failure. Every variant but the first is the worker rejecting the
+/// value; the setter leaves the chat as it was, and the worker keeps running.
+#[derive(Debug, thiserror::Error, miette::Diagnostic)]
 pub enum SetterError {
     #[error("Worker terminated before processing setter: {0}")]
     SetterError(String),
```
madsmtm: Nit: This could have a better name?

nobodywho/flutter/rust/src/lib.rs
```diff
@@ -670,16 +670,14 @@ impl RustChat {
...
             .await
-            .map_err(|e| e.to_string())
+            .map_err(|e| nobodywho::render_miette(&e))
     }
 
-    pub async fn set_sampler_config(
-        &self,
-        sampler_config: SamplerConfig,
-    ) -> Result<(), nobodywho::errors::SetterError> {
+    pub async fn set_sampler_config(&self, sampler_config: SamplerConfig) -> Result<(), String> {
```
madsmtm, line 676: Nice!

nobodywho/core/src/chat.rs
```diff
@@ -1426,19 +1441,17 @@ fn process_worker_msg(worker_state: &mut Chat<'_>, msg: ChatMsg) -> Result<(), C
...
+            let _ = output_tx.blocking_send(worker_state.set_tools(tools));
         }
         ChatMsg::SetSystemPrompt {
             system_prompt,
             output_tx,
         } => {
-            worker_state.set_system_prompt(system_prompt)?;
-            let _ = output_tx.blocking_send(());
+            worker_state.set_system_prompt(system_prompt);
+            let _ = output_tx.blocking_send(Ok(()));
```
madsmtm, line 1454: Remark: I kinda dislike that infallible methods like this become fallible (which is why I was in favour of the generic method discussed above), but I think it's fine for now.

nobodywho/python/src/lib.rs
```diff
@@ -8,6 +8,12 @@ use nobodywho::render_miette;
 
 mod parse;
 
+/// A setter the worker rejected. Rendered with the diagnostic help text, the
+/// way generation errors are.
+fn setter_err(e: nobodywho::errors::SetterError) -> PyErr {
```
madsmtm: Nit: I like this helper, but maybe you should make it more generic, and use it everywhere? There's still loads of `pyo3::exceptions::PyRuntimeError::new_err(render_miette(&e))` scattered around this file.

```suggestion
fn err(e: impl miette::Diagnostic) -> PyErr {
```

Can be done separately though.

madsmtm (approved):
4 sure the better approach

PR #702, opened 2026-08-31
fix: rejected inputs no longer kill the encoder worker

jona605a:
Encoding or ranking text longer than the context window propagated the
   error out of `process_worker_msg`, which ended the worker thread before it
   could answer on the caller's reply channel. The caller therefore saw only
   the "worker never responded" fallback, and the dead thread made every
   later `encode()` / `rank()` on that instance fail the same way.

   Both workers now carry the Result on the message's own reply channel, so
   the read error reaches its caller and the worker stays alive. Same shape
   as the chat setter fix in #701. `EncoderMsg` and `CrossEncoderMsg` are
   private and the public signatures already returned these error types, so
   no binding changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Format is based on [Keep a Changelog 1.1.0](https://keepachangelog.com/en/1.1.0/
 ### Fixed
 
 - A chat setter the worker rejects no longer kills the chat. Passing an invalid sampler config, or tools to a model with no tool-call template, ended the worker thread: the setter reported only "worker terminated", the real reason went to the log, and every later call on that chat — including `ask()` — failed the same way. The reason now reaches the caller and the chat keeps working, with the rejected value not applied. `set_chat_history` was fixed this way in the chat-completion change above; this covers `set_sampler_config`, `set_tools` and `reset_chat` too. Available for all bindings.
+- An input the encoder or cross-encoder rejects no longer kills it. Encoding or ranking text longer than the context window ended the worker thread, so the call reported only that the worker never responded, and every later `encode()` or `rank()` on that instance failed the same way. The reason — including the "input is too large for the context window" hint naming the token count and `n_ctx` — now reaches the caller, and the encoder stays usable for the next input. Available for all bindings.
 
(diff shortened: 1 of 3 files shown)


nobodywho/core/src/encoder.rs
```diff
@@ -72,18 +70,19 @@ impl EncoderAsync {
...
+        ))?
     }
 }
 
 enum EncoderMsg {
-    EncodeBatch(Vec<String>, tokio::sync::mpsc::Sender<Vec<Vec<f32>>>),
+    EncodeBatch(
+        Vec<String>,
+        tokio::sync::mpsc::Sender<Result<Vec<Vec<f32>>, EncoderWorkerError>>,
+    ),
```
madsmtm: Nit: Could we do:

```suggestion
    EncodeBatch {
        texts: Vec<String>,
        output_tx: tokio::sync::mpsc::Sender<Result<Vec<Vec<f32>>, EncoderWorkerError>>,
    },
```

To match `ChatMsg`?

Same for `CrossEncoderMsg`.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Format is based on [Keep a Changelog 1.1.0](https://keepachangelog.com/en/1.1.0/
 ### Fixed
 
 - A chat setter the worker rejects no longer kills the chat. Passing an invalid sampler config, or tools to a model with no tool-call template, ended the worker thread: the setter reported only "worker terminated", the real reason went to the log, and every later call on that chat — including `ask()` — failed the same way. The reason now reaches the caller and the chat keeps working, with the rejected value not applied. `set_chat_history` was fixed this way in the chat-completion change above; this covers `set_sampler_config`, `set_tools` and `reset_chat` too. Available for all bindings.
+- An input the encoder or cross-encoder rejects no longer kills it. Encoding or ranking text longer than the context window ended the worker thread, so the call reported only that the worker never responded, and every later `encode()` or `rank()` on that instance failed the same way. The reason — including the "input is too large for the context window" hint naming the token count and `n_ctx` — now reaches the caller, and the encoder stays usable for the next input. Available for all bindings.
```
madsmtm: Nit: Reads a bit sloppy, feels like both of these entries could be much shorter.

madsmtm (approved):

PR #703, opened 2026-08-31
docs/expo-docs

pielouNW:
## Description

Improve docs for expo support: 
- add expo in doc home page, navbar menu
- `Getting started` for expo in the docs with specific documentation
- mention Expo with React Native in readme
- add expo in llm.txt

Also improve nobodywho presence on npm/pub.dev/SPI/pypi (list all the features that were not presented + code snippets)

## Type of change

- [x] Documentation update

### DIFF
--- a/docs/docs-react-native/downloading-models.md
+++ b/docs/docs-react-native/downloading-models.md
@@ -1,6 +1,6 @@
 ---
 title: Downloading models
-description: How NobodyWho downloads, caches, and inspects GGUF models in React Native
+description: How NobodyWho downloads, caches, and inspects GGUF models in React Native & Expo
 sidebar_position: 1
 ---
 
@@ -18,7 +18,7 @@ The `modelPath` option to `Chat.fromPath` and `downloadModel` accepts:
 
 The HuggingFace prefix is case-insensitive and the `//` is optional — `hf:`, `hf://`, `huggingface:`, and `huggingface://` all mean the same thing. Remote models are downloaded to the platform cache directory on first load and re-used on subsequent runs.
 
-## Android permissions
+## Android permissions (React Native only)
 
 Loading a model from `hf://` or `https://` needs network access. The React Native app template already declares the internet permission in `android/app/src/main/AndroidManifest.xml`:
 
--- a/docs/docs-react-native/index.md
+++ b/docs/docs-react-native/index.md
@@ -1,6 +1,6 @@
 ---
 title: Getting started
-description: How to setup NobodyWho in React Native
+description: How to setup NobodyWho in React Native / Expo
 sidebar_position: 0
 ---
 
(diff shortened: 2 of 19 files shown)


docs/docs-react-native/downloading-models.md
```diff
@@ -18,7 +18,7 @@ The `modelPath` option to `Chat.fromPath` and `downloadModel` accepts:
 
 The HuggingFace prefix is case-insensitive and the `//` is optional — `hf:`, `hf://`, `huggingface:`, and `huggingface://` all mean the same thing. Remote models are downloaded to the platform cache directory on first load and re-used on subsequent runs.
 
-## Android permissions
+## Android permissions (React Native only)
```
madsmtm, line 21: Because Expo doesn't allow network access? Or because it automatically ensures it?

docs/react-native_versioned_docs/version-3.0.0/index.md
```diff
@@ -1,18 +1,52 @@
 ---
```
madsmtm, line 1: Remark: Dunno what the plan is for old docs? I'd prefer to have the policy that we don't update them, but perhaps you think otherwise?

docs/docs-react-native/index.md
```diff
@@ -1,18 +1,52 @@
...
 
+### React Native
+
 No additional initialization step is required — the native module is loaded automatically when you first import from the package.
 
+### Expo
+
+NobodyWho ships native code, so it does **not** run in [Expo Go](https://docs.expo.dev/get-started/set-up-your-environment/). You need a [development build](https://docs.expo.dev/develop/development-builds/introduction/).
+
+#### • Continuous Native Generation (CNG) projects
```
madsmtm, line 25: Nit: Remove weird char? Unless there was a point to it?

```suggestion
#### Continuous Native Generation (CNG) projects
```

Same below.

docs/docs-react-native/index.md
```diff
@@ -1,18 +1,52 @@
...
+
+In a managed project, `ios/` and `android/` are not committed. For the **first build**, run the command for the platform you are targeting (`run:ios` and `run:android` are alternatives — pick your target). It runs prebuild automatically since the native folders don't exist yet, installs pods, and builds NobodyWho in:
```
madsmtm, line 27: Nit: Only installs pods on iOS, the setup step on Android is different.

docs/docs-react-native/index.md
```diff
@@ -1,18 +1,52 @@
...
+```bash
+npx expo run:ios
+# or
+npx expo run:android
+```
+
+After you **upgrade** NobodyWho (or any other native dependency), the native folders already exist, so `run:*` would build against stale native code. Regenerate them from scratch, then rebuild:
+
+```bash
+npx expo prebuild --clean   # regenerate ios/ and android/ from scratch
```
madsmtm, line 38: It feels like it'd be easier to just always say "run prebuild clean", regardless of whether you just added `nobodywho` as a dependency or are upgrading it? Or does it explicitly not work if you do it in that order?

nobodywho/flutter/nobodywho/pubspec.yaml
```diff
@@ -9,8 +9,8 @@ homepage: https://docs.nobodywho.ooo/
 topics:
   - ai
   - llm
-  - genai
-  - inference
```
madsmtm, line 13: Remark: Inference is a pretty important tag IMO. But dunno, I guess LLM and AI are enough to capture people searching for that.

nobodywho/swift/README.md
```diff
@@ -180,6 +180,55 @@ let chat = try await Chat.fromPath(
 // SamplerPresets.constrainWithGrammar("...")        — constrain to a Lark or GBNF grammar
 ```
 
+### Speech to Text
+
+Transcribe spoken audio into text using Whisper models in ONNX format:
+
+```swift
```
madsmtm, line 187: Question: Do we have testing anywhere of our README examples? At least testing that it compiles / doesn't give a syntax error?

madsmtm (approved):
Nice that it works, and definitely good to have this documented!

docs/docs-react-native/downloading-models.md
pielouNW: good catch. it's automatically added by expo

docs/react-native_versioned_docs/version-3.0.0/index.md
pielouNW: Update if we can, but up to the rest of the team imo

docs/docs-react-native/index.md
pielouNW: Yes, it's to visually divide the 2 approach

docs/docs-react-native/index.md
pielouNW: Some people might have changes in /ios or /android, prebuild clean would wipe these changes

nobodywho/flutter/nobodywho/pubspec.yaml
pielouNW: agreed, it's important, but we are limited in amount of topics keywords
it's to capture stt / tts / local-ai topics
mobile devs are more aware of these terms than inference / gen-ai 
up to debate

nobodywho/swift/README.md
pielouNW: Not that I am aware

docs/docs-react-native/index.md
madsmtm: Fair enough

docs/docs-react-native/index.md
madsmtm: Ah, makes sense. What about just `npx expo prebuild`, without cleaning? That wouldn't work after updating the dep?

PR #705, opened 2026-09-01
perf: reuse the llguidance parser factory for tool grammars

jona605a:
## Description

Since #674, `Chat::apply_options` called `set_sampler_config` then `set_tools`, and
both rebuilt the tool grammar — so a `complete(messages,options=Options(sampler, tools))` compiled it
twice. `set_sampler_config` also recompiled it at all, though the grammar depends
only on the tools.

Setting tools go through this "flow":
`tools: &[Tool]` becomes `lark: String` (matching the model's format) becomes `ShiftStep` becomes `LlamaSampler` (just the grammar shift step) becomes a chained `LlamaSampler` with all shiftsteps and a sample step in the end. 

The compilation to a sampler consists of going through the model's `TokEnv` and using it to create a `ParserFactory` (the two expensive steps) which are then combined with the specific tool's lark grammar to create a sampler (cheap). This PR caches the ParserFactory in Chat as `Option<GrammarFactory>`.

`Chat::apply_sampler_and_tools(Option<SamplerConfig>, Option<Vec<Tool>>)` — one path for `set_sampler_config`, `set_tools` and `apply_options`; each sampler built once, all-or-nothing on failure. Empty `Options` does no sampler work.

Measured, Qwen3-0.6B Q4_K_M, three turns setting both sampler and tools:

| | before | after |
|---|---|---|
| worker setup | ~420 ms | ~430 ms |
| per later rebuild | **805 ms** | **6 ms** |

Steady-state memory is unchanged: a live tool sampler's `TokenParser` already
pinned the `TokEnv` and slicer via `Arc` (~29 MiB measured on the pre-PR path), so
this removes rebuild churn, not footprint.


nobodywho/core/src/chat.rs
```diff
@@ -2365,17 +2371,70 @@ impl<'a> Chat<'a> {
...
+            &mut self.grammar_factory,
+            effective_tools,
+            config,
             self.tool_format.as_ref(),
         )?;
-        self.sampler_config = sampler_config;
-        self.sampler.set_base(base_sampler);
+        // Only a tool change re-selects the template; doing it on a sampler
+        // change would alter the render and force a re-prefill.
+        let chat_template = match &tools {
```
s194042, line 2421: In theory, someone could pass a list of tools that is identical to the ones already loaded, which would make this step unnecessary. Maybe too niche of a case, but mentioning anyways

s194042 (approved):

nobodywho/core/src/chat.rs
jona605a: True, fixing that. The more time-consuming part of that case, though, would be to recompute the parser from the ParserFactory. It's only ~6 ms on my machine. Should we key on the lark string of the tools?

PR #706, opened 2026-09-01
Fix building Android Flutter package on Windows

### LINKED ISSUE #437: [BUG] Create a new blank project, but it cannot be run.A problem occurred starting process 'command 'dart''
```
Execution failed for task ':nobodywho:resolveNativeLibraries'.
> A problem occurred starting process 'command 'dart''
```
Image

Image

madsmtm:
We previously tried to invoke just `dart`, now we try to invoke `dart.exe` (from the `FLUTTER_ROOT`).

Pretty sure this fixes https://github.com/nobodywho-ooo/nobodywho/issues/437, cc @peng093 @Trung15010802.

(Note that I know very little about both Gradle and Flutter, there probably exists a better way to do this).

### DIFF
--- a/nobodywho/flutter/nobodywho/android/build.gradle.kts
+++ b/nobodywho/flutter/nobodywho/android/build.gradle.kts
@@ -8,6 +8,7 @@
 // 4. Download from GitHub releases
 
 import java.io.ByteArrayOutputStream
+import java.io.File
 import org.gradle.process.ExecOperations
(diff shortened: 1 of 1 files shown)


jona605a (approved):
Also not an expert on gradle / flutter, but this seems good

madsmtm:
Ooops, should've run that before requesting a review

PR #707, opened 2026-09-01
Add experimental OpenAI-compatible server

duarteocarmo:
- Experimental OpenAI Chat Completions server with streaming and tool calls.
- Adds minimal, unlisted documentation with Pi integration instructions.

### DIFF
--- a/nobodywho/Cargo.lock
+++ b/nobodywho/Cargo.lock
@@ -274,6 +274,58 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
 
 [[package]]
+name = "axum"
+version = "0.8.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "31b698c5f9a010f6573133b09e0de5408834d0c82f8d7475a89fc1867a71cd90"
+dependencies = [
+ "axum-core",
+ "bytes",
+ "form_urlencoded",
+ "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
+ "hyper",
+ "hyper-util",
+ "itoa",
+ "matchit",
+ "memchr",
+ "mime",
+ "percent-encoding",
+ "pin-project-lite",
+ "serde_core",
+ "serde_json",
+ "serde_path_to_error",
+ "serde_urlencoded",
+ "sync_wrapper",
+ "tokio",
+ "tower",
+ "tower-layer",
+ "tower-service",
+ "tracing",
+]
+
+[[package]]
+name = "axum-core"
+version = "0.5.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08c78f31d7b1291f7ee735c1c6780ccde7785daae9a9206026862dab7d8792d1"
+dependencies = [
+ "bytes",
+ "futures-core",
+ "http",
+ "http-body",
+ "http-body-util",
+ "mime",
+ "pin-project-lite",
+ "sync_wrapper",
+ "tower-layer",
+ "tower-service",
+ "tracing",
+]
+
+[[package]]
 name = "backtrace"
 version = "0.3.76"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2069,6 +2121,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6dbf3de79e51f3d586ab4cb9d5c3e2c14aa28ed23d180cf89b4df0454a69cc87"
 
 [[package]]
+name = "httpdate"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "df3b46402a9d5adb4c86a0cf463f42e19994e3ee891101b1841f30a545cb49a9"
+
+[[package]]
 name = "hyper"
 version = "1.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2082,6 +2140,7 @@ dependencies = [
  "http",
  "http-body",
  "httparse",
+ "httpdate",
  "itoa",
  "pin-project-lite",
  "smallvec",
@@ -2844,6 +2903,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "matchers"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d1525a2a28c7f4fa0fc98bb91ae755d1e2d1505079e05539e35bc876b5d65ae9"
+dependencies = [
+ "regex-automata",
+]
+
+[[package]]
+name = "matchit"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47e1ffaa40ddd1f3ed91f717a33c8c0ee23fff369e3aa8772b9605cc1d22f4c3"
+
+[[package]]
 name = "matrixmultiply"
 version = "0.3.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2920,6 +2994,12 @@ dependencies = [
  "quote",
  "syn 2.0.119",
 ]
+
+[[package]]
+name = "mime"
+version = "0.3.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6877bb514081ee2a7ff5ef9de3281f14a4dd4bceac4c09388074a6b5df8a139a"
 
 [[package]]
 name = "minijinja"
@@ -3195,6 +3275,21 @@ dependencies = [
  "rand 0.9.5",
  "serde_json",
  "tokio",
+ "tracing",
+ "tracing-subscriber",
+]
+
+[[package]]
+name = "nobodywho-server"
+version = "0.1.0"
+dependencies = [
+ "axum",
+ "clap",
+ "nobodywho",
+ "serde",
+ "serde_json",
+ "tokio",
+ "tokio-stream",
  "tracing",
  "tracing-subscriber",
 ]
@@ -4671,6 +4766,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_path_to_error"
+version = "0.1.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "10a9ff822e371bb5403e391ecd83e182e0e77ba7f6fe0160b795797109d1b457"
+dependencies = [
+ "itoa",
+ "serde",
+ "serde_core",
+]
+
+[[package]]
 name = "serde_spanned"
 version = "0.6.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4686,6 +4792,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6662b5879511e06e8999a8a235d848113e942c9124f211511b16466ee2995f26"
 dependencies = [
  "serde_core",
+]
+
+[[package]]
+name = "serde_urlencoded"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d3491c14715ca2294c4d6a88f15e84739788c1d030eed8c110436aafdaa2f3fd"
+dependencies = [
+ "form_urlencoded",
+ "itoa",
+ "ryu",
+ "serde",
 ]
 
 [[package]]
@@ -5472,6 +5590,7 @@ dependencies = [
  "tokio",
  "tower-layer",
  "tower-service",
+ "tracing",
 ]
 
 [[package]]
@@ -5554,10 +5673,14 @@ version = "0.3.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cb7f578e5945fb242538965c2d0b04418d38ec25c79d160cd279bf0731c8d319"
 dependencies = [
+ "matchers",
  "nu-ansi-term",
+ "once_cell",
+ "regex-automata",
  "sharded-slab",
  "smallvec",
  "thread_local",
+ "tracing",
  "tracing-core",
  "tracing-log",
 ]
--- a/nobodywho/Cargo.toml
+++ b/nobodywho/Cargo.toml
@@ -1,6 +1,17 @@
 [workspace]
 members = [
     "core",
+    "server",
+    "godot",
+    "flutter/rust",
+    "python",
+    "grammar/gbnf",
+    "grammar/gbnf-macro",
+    "uniffi",
+]
+default-members = [
+    "core",
+    "server",
     "godot",
     "flutter/rust",
     "python",
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -945,6 +945,28 @@ impl ChatHandleAsync {
         )))
     }
 
+    /// Generate one assistant response and return tool calls to the caller.
+    ///
+    /// Unlike [`complete`](Self::complete), this never executes tool callbacks.
+    /// The caller can execute returned calls and submit the full conversation,
+    /// including the tool results, in a following request.
+    pub fn complete_with_external_tools(
+        &self,
+        messages: Vec<Message>,
+        options: Options,
+        max_tokens: Option<usize>,
+    ) -> Result<CompletionStreamAsync, InvalidHistoryError> {
+        validate_completion_messages(&messages)?;
+        let (output_tx, output_rx) = tokio::sync::mpsc::channel(32);
+        self.guard.send(ChatMsg::CompleteWithExternalTools {
+            messages,
+            options,
+            max_tokens,
+            output_tx,
+        });
+        Ok(CompletionStreamAsync::new(output_rx))
+    }
+
     /// Send a setter message and wait for the worker's answer. The outer error
     /// is the worker being gone, the inner one is the worker rejecting the value.
     async fn set_and_wait_async<F>(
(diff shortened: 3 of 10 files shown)


nobodywho/core/src/lib.rs
```diff
@@ -44,6 +44,10 @@ pub fn send_llamacpp_logs_to_tracing() {
     llama_cpp_2::send_logs_to_tracing(llama_cpp_2::LogOptions::default().with_logs_enabled(true));
 }
 
+pub fn suppress_llamacpp_logs() {
+    llama_cpp_2::send_logs_to_tracing(llama_cpp_2::LogOptions::default().with_logs_enabled(false));
+}
```
madsmtm: The better option for this is probably to filter the logs you don't want out? Something like:
```rust
tracing_subscriber::fmt().with_env_filter(
    EnvFilter::new("llama.cpp=trace,ggml=trace")
).init();
```

(I'm a little bit wary of using `llama_cpp_2::send_logs_to_tracing` too much, because it's actually `unsafe`).

nobodywho/server/pyproject.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+description = "OpenAI-compatible server for NobodyWho"
+requires-python = ">=3.10"
+license = { text = "EUPL-1.2" }
+
+[project.urls]
+Repository = "https://github.com/nobodywho-ooo/nobodywho"
+
+[tool.maturin]
+bindings = "bin"
```
madsmtm, line 16: This is for `uvx`? Could we make it possible to easily use `cargo install --git ...` instead? I guess that's less easy than `uvx` though, since you have to manually handle where it's installed and such.

nobodywho/Cargo.lock
```diff
@@ -284,6 +284,58 @@ version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
 
+[[package]]
```
madsmtm, line 276: Cargo.nix should be updated as well IIUC? Though why didn't CI catch that?

nobodywho/Cargo.toml
```diff
@@ -1,5 +1,15 @@
...
+    "core",
+    "server",
+    "godot",
+    "flutter/rust",
+    "python",
+    "grammar/gbnf",
+    "grammar/gbnf-macro",
+    "uniffi",
+]
+default-members = [
```
madsmtm, line 12: You don't want the server as part of the default members? Why not? `axum` isn't that big of a dependency in the grand scheme of things, so compiling that by default is probably not a huge issue?

nobodywho/server/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "nobodywho-server"
+version = "0.1.0"
+edition = "2021"
+
+[[bin]]
+name = "nobodywho-server"
+path = "src/main.rs"
```
madsmtm: Nit: this shouldn't be necessary

nobodywho/core/src/chat.rs
```diff
@@ -1925,9 +2047,21 @@ impl<'a> Chat<'a> {
     // contexts with the same model. It might not be necessary
     // but assume it is.
     pub fn generate_response_until_done<F>(
```
madsmtm: Nit: Just inline this in `wrapped_update_context_and_generate_response`.

nobodywho/core/src/tool_calling/mod.rs
```diff
@@ -322,14 +322,20 @@ pub(crate) fn lark_delimiter(model: Option<&LlamaModel>, s: &str) -> String {
...
 }
 
-/// Map any non-alphanumeric character to `_` for use in Lark rule names.
+/// Map input to lowercase alphanumeric characters and `_` for Lark rule names.
 ///
 /// Tool/property names come from user-controlled input and are spliced into
-/// generated Lark rule names, which only allow a restricted identifier
-/// charset.
+/// generated Lark rule names. Lark treats uppercase identifiers as terminals,
+/// so generated rule identifiers must remain lowercase.
```
madsmtm, line 329: Nice find!

nobodywho/core/src/chat.rs
```diff
@@ -944,6 +944,28 @@ impl ChatHandleAsync {
         )))
     }
 
+    /// Generate one assistant response and return tool calls to the caller.
+    ///
+    /// Unlike [`complete`](Self::complete), this never executes tool callbacks.
+    /// The caller can execute returned calls and submit the full conversation,
+    /// including the tool results, in a following request.
+    pub fn complete_with_external_tools(
```
madsmtm, line 953: Hmm, I kinda dislike all this extra ceremony, I feel like there could be an easier way of handling this? It just seems like there's a lot of duplication between `run_turn`/`run_turn_once` and `complete`/`complete_once` etc.?

Maybe we can implement `complete` on top of `complete_with_external_tools`? Or maybe just have a flag `skip_calling_tools` that is checked in `run_turn`?

Does that make sense, or WDYT?

nobodywho/core/src/inference.rs
madsmtm, line 475: Question: Why doesn't this also need a `max_output_tokens: Option<usize>` parameter?

nobodywho/core/src/inference.rs
```diff
@@ -522,7 +526,10 @@ impl<'a> InferenceEngine<'a> {
         // Clamp drafts so the verify batch [pending, drafts...] stays
         // within the context window:
         let room = usize::try_from(self.ctx.n_ctx() as i32 - self.n_past - 1).unwrap_or(0);
-        drafts.truncate(room);
+        let output_room = max_output_tokens
+            .map(|max_output_tokens| max_output_tokens.saturating_sub(1))
+            .unwrap_or(usize::MAX);
+        drafts.truncate(room.min(output_room));
```
madsmtm, line 532: `max_output_tokens` is intended as a poor man's [`ContextOverflow`](https://app.notion.com/p/Chat-Completion-API-3b5f845da7068078951ec7bf236a32e8#3c2f845da70680f3b8a7e981ca7b80c1), right?

nobodywho/server/src/main.rs
```diff
@@ -0,0 +1,1191 @@
+use axum::extract::rejection::JsonRejection;
```
madsmtm, line 1: Could you add some docs from Pi or OpenAI or similar on what the expected HTTP API we're implementing is actually is?

Just for those of us unaware, it'd be nice to know "ah, `/v1/chat/completions` uses SSE when the `stream` parameter is set".

nobodywho/server/src/main.rs
```diff
@@ -0,0 +1,1191 @@
...
+        _ = terminate.recv() => {},
+    }
+}
+
+#[cfg(not(unix))]
+async fn shutdown_signal() {
+    tokio::signal::ctrl_c()
+        .await
+        .expect("failed to install Ctrl-C handler");
+}
```
madsmtm, line 319: Remark: This is sick! Tokio is hella weird!

nobodywho/server/src/main.rs
```diff
@@ -0,0 +1,1191 @@
...
+
+    for message in messages {
+        match message {
+            RequestMessage::System { content } | RequestMessage::Developer { content }
+                if !saw_conversation =>
+            {
+                system_parts.push(content.into_text().map_err(bad_request)?);
+            }
+            RequestMessage::System { .. } | RequestMessage::Developer { .. } => {
+                return Err(bad_request("system and developer messages must come first"));
```
madsmtm, line 436: CC @jona605a, here's an example where allowing (in our model) `Message::System` to be anywhere would be useful for developers (even if it later throws an error).

nobodywho/server/src/main.rs
```diff
@@ -0,0 +1,1191 @@
...
+    max_tokens: usize,
+}
+
+#[derive(Deserialize)]
+struct ChatCompletionRequest {
+    model: String,
+    messages: Vec<RequestMessage>,
+    #[serde(default)]
+    tools: Vec<ToolDefinition>,
+    tool_choice: Option<Value>,
```
madsmtm: Nit: This could probably be parse up-front with serde too

nobodywho/server/src/main.rs
```diff
@@ -0,0 +1,1191 @@
...
+    context_window: u32,
+    max_tokens: usize,
+}
+
+#[derive(Deserialize)]
+struct ChatCompletionRequest {
+    model: String,
+    messages: Vec<RequestMessage>,
+    #[serde(default)]
+    tools: Vec<ToolDefinition>,
```
madsmtm: Nit: I think you could use `#[serde(deserialize_with = ...)]` to go to `Vec<Tool>` here directly (in the vein of "parse, don't validate").

nobodywho/server/src/main.rs
```diff
@@ -0,0 +1,1191 @@
...
+    let model =
+        Arc::new(llm::get_model_async(config.model.clone(), !config.cpu, None, None, None).await?);
+    let context_window = config.context_size.min(model.max_ctx());
+    let mut builder = ChatBuilder::new(model).with_context_size(context_window);
+    if let Some(threads) = config.threads {
+        builder = builder.with_n_threads(threads);
+    }
+    let chat = builder.build_async()?;
+    let default_sampler = chat.get_sampler_config().await?;
+    let model_id = config.name.unwrap_or_else(|| model_id_for(&config.model));
```
madsmtm, line 273: Remark: This seems brittle? We should probably allow the user to get more data out of the `gguf`!

nobodywho/server/src/main.rs
```diff
@@ -0,0 +1,1191 @@
...
+        let completion_request = json!({
+            "model": self.model.id,
+            "messages": [{"role": "user", "content": "Say hello in one short sentence."}],
+            "stream": true,
+        });
+        write!(
+            formatter,
+            "Server ready\n\n  Address:  {base_url}\n  Model ID: {}\n  Context:  {} tokens\n\n  Routes:\n{routes}\n\n  Try it:\n    curl {base_url}{HEALTH_PATH}\n    curl {base_url}{MODELS_PATH}\n    curl -N {base_url}{CHAT_COMPLETIONS_PATH} \\\n      -H 'Content-Type: application/json' \\\n      -d '{completion_request}'",
+            self.model.id, self.model.context_window
+        )
```
madsmtm: Nit: Split across multiple `writeln!(f, "...")?;` instead.

nobodywho/server/src/main.rs
madsmtm: Wait, I realize now that you have to use a custom Pi extension:
https://gist.github.com/duarteocarmo/39a3f1148e2e33c27fddbbef2d971deb

Can we avoid that somehow?

madsmtm (changes requested):
I like the idea, especially also that we have some bigger example that utilizes / stress-tests the `core` API!
For example, we talked today about more structured content output, and having the server implement exactly that would allow more easily seeing how having it in `core` would simplify the server.

Only blocking concern is the duplication and extra complexity in `chat.rs`.

As a general remark, feels like this PR is doing a bit too much, I'd have preferred for it to be split into like three PRs (one adding `max_output_tokens`, one modifying `chat.rs` with `CompleteWithExternalTools`, and the last one adding the server).

nobodywho/server/src/main.rs
duarteocarmo: For using it with Pi, not really. We can't avoid it. But it's just an example - with an openai compatible server - the world is your oyster !

nobodywho/server/pyproject.toml
duarteocarmo: Yeah this is for uvx. The idea here is that I wanted to have a single command that spins everything up. No install required. I just think it makes for a better ux.

nobodywho/Cargo.toml
duarteocarmo: Didn't want to pollute with something unreleased. But can add back.

nobodywho/server/src/main.rs
duarteocarmo: 😂 I agree.

nobodywho/server/src/main.rs
madsmtm: Fair enough, makes sense. Do you want to maintain the script as part of this repo too then? Or is it basically never going to need to be updated?

duarteocarmo: Yeah I thought about that. But pi doesn't allow me to point at this repo with pi -e git:... unless I put a package.json file on the root of the repo, which I didn't like. 

The alternative is to create *another* repo - but that seems excessive. So this was the best "keep it hidden" - dont pollute alternative I could come up with. 

If we make the server something user facing, we could def. think about putting somewhere more maintainable. 

thoughts?

nobodywho/server/pyproject.toml
duarteocarmo: Easy is the name of the game :)

nobodywho/core/src/chat.rs
duarteocarmo: > Or maybe just have a flag skip_calling_tools that is checked in run_turn?

Like this, went with it.

nobodywho/core/src/inference.rs
duarteocarmo: Solo inference produces a single token. The `chat.rs` enforces the `max_tokens` in our normal chat inference (by looping over the `sample_and_decode_solo`

nobodywho/core/src/inference.rs
duarteocarmo: Yeah, pretty much. Most LLM apis allow for a max_output_tokens and we enforce it here in the case of speculative..

nobodywho/server/src/main.rs
duarteocarmo: Not sure what you mean here?

duarteocarmo:
@madsmtm agree with splitting stuff. I'll for sure do it next time! Or maybe use stacked PRs. I think I've addressed everything - but let me know if not!

nobodywho/server/src/main.rs
AsbjornOlling: Could be about metadata. Was also requested by a user here:
https://github.com/nobodywho-ooo/nobodywho/pull/658#issuecomment-5232369023

duarteocarmo: So that the openai server exposes it? Which endpoint?

nobodywho/server/src/main.rs
madsmtm: Maybe leave it as a FIXME or TODO in the docs?

Also, I'd still like a link to https://developers.openai.com/api/reference/overview somewhere in here.

nobodywho/server/src/main.rs
madsmtm: Yeah, I meant that we should probably expose gguf metadata (`llama_cpp_2::gguf::GgufContext::from_file`) somehow in the future.

(Btw, I prefixed this with `Remark`, to try to signal "this doesn't need to be fixed now, it's just something that I thought of". Idk., it's not the best system).

nobodywho/core/src/chat.rs
madsmtm: Hmm, I thought you could maybe get rid of `ChatMsg::CompleteWithExternalTools` as well if you did it that way, but I see that it's difficult, since that returns a single value while `ChatMsg::Complete` returns multiple.

And I guess we want to continue running tools on the worker thread.

Let's say it's good enough for now, but I definitely want to refactor this at some point.

nobodywho/core/src/chat.rs
```diff
@@ -2202,11 +2354,32 @@ impl<'a> Chat<'a> {
 
         self.hoist_system_message(&mut messages)?;
         self.messages = messages;
-        self.run_turn(respond)?;
+        self.run_turn(respond, None, false)?;
 
         Ok(self)
     }
 
+    fn complete_once<F>(
```
madsmtm, line 2385: Nit: You could probably merge `complete` and `complete_once` as well by adding all the parameters from `run_turn` to it?

nobodywho/core/src/chat.rs
```diff
@@ -2303,26 +2476,26 @@ impl<'a> Chat<'a> {
...
             .tool_format
             .as_ref()
-            .map(|fmt| fmt.begin_token().to_string());
-
-        // wrap the response callback to keep a copy of the completed response
-        // and to avoid emitting tool calls
+            .map(|format| format.begin_token().to_string());
         let (wrapped_respond, resp_receiver) =
-            crate::inference::wrap_respond(respond.clone(), tool_call_begin_token);
+            crate::inference::wrap_respond(respond, tool_call_begin_token);
```
madsmtm, line 2497: 🔫 that clone!

nobodywho/core/src/chat.rs
duarteocarmo: Agreed!

nobodywho/server/src/main.rs
duarteocarmo: Agreed - will leave as is for now.

madsmtm (approved):
> agree with splitting stuff. I'll for sure do it next time! Or maybe use stacked PRs.

I think either would be a good idea. Not sure how good GH's support for stacked PRs is yet, but yeah, we could try that out too.

> I think I've addressed everything - but let me know if not!

I think it's good as well, feel free to merge when https://github.com/nobodywho-ooo/nobodywho/pull/707#discussion_r3924576226 is resolved (the rest is just nits).

nobodywho/server/pyproject.toml
madsmtm: I agree, am fine with keeping this

nobodywho/Cargo.toml
madsmtm: Thanks!

As a nit: You can also remove the `default-members` now, since it's the same as `members`.

PR #709, opened 2026-09-04
feat: System messages anywhere

jona605a:
In #674, I changed the system prompt from sitting at `messages[0]` to being a setting on `Chat`, and then disallowed any other system messages in the chat history. 

However, some models allow mid-conversation system messages, so it would be nice to support that. 

For the chat templates that don't allow system messages, instead of encountering the error during generation and being stuck on a bad system history, I do a test-render in `complete` and `set_chat_history`. Not sure if there's a better solution, but this should be cheap enough.

For context shifting, I chose to always keep system messages, as it should be unlikely that more than half of the context is filled up with system messages. However, in the future, it would be better to support a `ContextOverflow` option where users can configure context shift policies more.


marek-hradil (changes requested):
Just get rid of the `check_...` functions, as we talked:
- creating `Messages` struct, which will accept the current array and perform the validations, so that its impossible to create `Messages` without them being correct.

jona605a:
Regarding the ai review comment: as discussed with Marek, I removed the test-render and just accepted that if the history contains a system message that is unrenderable, then we accept the error

marek-hradil (approved):

PR #711, opened 2026-09-05
Compatible completions (1/3): Core metadata and sampling

duarteocarmo:
- Add request-level sampling and output limits to the core completion path.
- Return finish reasons and token usage with buffered and streamed completions.
- Reuse the existing `Options` type and keep output limits as an explicit argument.
- Update the HTTP server to use the shared completion result.


nobodywho/core/src/chat.rs
```diff
@@ -250,6 +250,80 @@ impl Options {
...
+fn validate_sampling_overrides(
+    temperature: Option<f32>,
+    top_p: Option<f32>,
+) -> Result<(), SamplingOverrideError> {
+    if temperature.is_some_and(|value| !(0.0..=2.0).contains(&value)) {
+        return Err(SamplingOverrideError::InvalidTemperature);
+    }
+    if top_p.is_some_and(|value| !(0.0..=1.0).contains(&value)) {
+        return Err(SamplingOverrideError::InvalidTopP);
+    }
```
madsmtm: Q: Why are these checks necessary? Can't we just allow this?

nobodywho/core/src/chat.rs
```diff
@@ -250,6 +250,80 @@ impl Options {
...
+        if temperature != Some(0.0) && matches!(&sampler.sample_step, SampleStep::Greedy) {
+            sampler.sample_step = SampleStep::Dist;
+        }
+        replace_or_push_sampling_step(
+            &mut sampler.steps,
+            |step| matches!(step, ShiftStep::TopP { .. }),
+            ShiftStep::TopP { top_p, min_keep: 1 },
+        );
+    }
+    Ok(sampler)
```
madsmtm, line 326: See https://github.com/nobodywho-ooo/nobodywho/pull/712#discussion_r3989647157, I think we'd be able to get rid of this logic if we did that.

I regret doing this as two PRs now... 😬

nobodywho/core/src/chat.rs
```diff
@@ -2498,13 +2777,21 @@ impl<'a> Chat<'a> {
         let (wrapped_respond, resp_receiver) =
             crate::inference::wrap_respond(respond, tool_call_begin_token);
 
-        self.generate_response_until_done_with_limit(
+        let prompt_tokens = self.context.chunks.n_tokens();
+        let available_tokens = (self.engine.ctx.n_ctx() as usize).saturating_sub(prompt_tokens);
+        let generation_limit = max_tokens.map(|limit| min(limit, available_tokens));
```
madsmtm, line 2782: Remark: It feels sorta weird to override the user's requested limit.

Is it because you want to completely avoid the context shift if a `max_tokens` limit is set?

nobodywho/core/src/chat.rs
```diff
@@ -4710,6 +4997,39 @@ mod tests {
...
+            .any(|step| matches!(step, ShiftStep::TopP { top_p, .. } if *top_p == 0.8)));
+        Ok(())
+    }
+
+    #[test]
+    #[allow(
+        clippy::type_complexity,
+        reason = "The exact public function signature is what this test checks"
+    )]
+    fn async_external_tools_signature_is_compatible() {
```
madsmtm, line 5022: Nit: Why do you want to test this?

nobodywho/core/src/chat.rs
```diff
@@ -2299,21 +2585,16 @@ impl<'a> Chat<'a> {
...
 
                 for tool_call in tool_calls {
-                    // find the tool
-                    // this is just a stupid linear search
-                    // but I think it's probably faster than something fancy as long as we have few tools
-                    // /shrug I'm happy to be wrong
                     let Some(tool) = self.tools.iter().find(|t| t.name == tool_call.name) else {
-                        // in case the tool isn't found.
-                        // I *think* this should be impossible, as long as the tool calling grammar
-                        // works.
```
madsmtm, line 2316: I tend to think these comments are somewhat valuable.

Or maybe not the `find the tool` and `in case the tool isn't found`, but the rest are?

nobodywho/core/src/chat.rs
```diff
@@ -1306,6 +1466,110 @@ impl CompletionStreamAsync {
...
+        &mut self,
+    ) -> Result<Option<StructuredCompletionChunk>, crate::errors::CompletionError> {
+        self.0.next_async().await.map(|output| {
+            output.map(|output| match output {
+                CompletionOutput::Token(token) => StructuredCompletionChunk::Token(token),
+                CompletionOutput::Done(response) => StructuredCompletionChunk::Done(response),
+            })
+        })
+    }
 }
```
madsmtm, line 1573: Could you explain what these are for? I don't immediately understand it.

madsmtm (commented):
Gave it a quick look, will look in more detail on Monday

duarteocarmo:
@madsmtm  thanks for the comments - I think makes more sense as a single PR - will follow up on this soon!

nobodywho/core/src/chat.rs
duarteocarmo: The old stream only returns content and tool calls. We now use these wrappers in the new API to send the full result back to the caller. 

e.g., 

 ```text
   Worker sends the FULL result
     Token("Hello")
     Token("!")
     Done {
       content,
       tool_calls,
       finish_reason,
       usage
     }
           │
           ▼
   CompletionReceiver
     internal pipe reader
     does not remove anything
           │
           ├── old public wrapper
           │   CompletionStreamAsync
           │     └── removes finish_reason + usage
           │
           └── new public wrapper
               StructuredCompletionStream
                 └── keeps the full result
 ```

nobodywho/core/src/chat.rs
duarteocarmo: I put them back.

nobodywho/core/src/chat.rs
duarteocarmo: It can still context shift before we get there. But once we get close enough I would like to avoid - unless you have a better idea :)

nobodywho/core/src/chat.rs
duarteocarmo: You're right - useless - removed. (was just to confirm signature stayed the same)

duarteocarmo:
@madsmtm ready for another round!

nobodywho/core/src/chat.rs
```diff
@@ -1359,10 +1549,11 @@ enum ChatMsg {
         options: Options,
         output_tx: tokio::sync::mpsc::UnboundedSender<llm::WriteOutput>,
     },
-    CompleteWithExternalTools {
+    StructuredComplete {
         messages: Vec<Message>,
         options: Options,
         max_tokens: Option<usize>,
+        execute_tools: bool,
```
madsmtm, line 1616: Nit: Let's use the same terminology in `complete_once`/`run_turn`.

(So either `skip_calling_tools` everywhere or `execute_tools` everywhere, I don't have a strong opinion on either).

nobodywho/core/src/chat.rs
```diff
@@ -2167,10 +2374,12 @@ impl<'a> Chat<'a> {
             respond(WriteOutput::Token(token_str));
         }
 
+        hit_token_limit |= max_tokens.is_some_and(|max_tokens| generated_tokens >= max_tokens);
```
madsmtm: Nit: This can be calculated outside of `generate_response_until_done_with_limit` (i.e. inside `wrapped_update_context_and_generate_response`) from just `generated_tokens` and `max_tokens`, I'd prefer to do that (keeps the logic closer together).

(So just return `generated_tokens` from this method)

nobodywho/core/src/chat.rs
madsmtm: I'd prefer if we removed the old wrapper then, I don't see much benefit to keeping it?

E.g. merge `AssistantResponse`/`CompletionResponse` and merge `StructuredCompletionChunk`/`CompletionChunk`/`CompletionOutput` etc.?

nobodywho/core/src/chat.rs
```diff
@@ -959,15 +979,28 @@ impl ChatHandleAsync {
...
             options,
             max_tokens,
-            output_tx,
-        });
-        Ok(CompletionStreamAsync::new(output_rx))
+            false,
+        )?))
+    }
+
+    pub fn complete_with_external_tools_and_metadata(
```
madsmtm: Nit: It seems like we now have:
- `ChatHandle::complete_with_metadata`
- `ChatHandle::complete_with_external_tools`
- `ChatHandleAsync::complete_with_external_tools`
- `ChatHandleAsync::complete_with_external_tools_and_metadata`

That seems inconsistent?

nobodywho/core/src/chat.rs
```diff
@@ -1254,48 +1287,101 @@ pub type TokenStream = crate::stream::TokenStream<crate::errors::CompletionError
...
-/// An async completion stream that returns tool calls instead of executing them.
-pub struct CompletionStreamAsync {
+enum CompletionOutput {
+    Token(String),
+    Done(CompletionResponse),
+}
+
+struct CompletionReceiver {
     rx: tokio::sync::mpsc::Receiver<ExternalCompletionOutput>,
+    _guard: Arc<WorkerGuard<ChatMsg>>,
```
madsmtm: This is subtle, and I fear that it might not be the right thing? This keeps the chat alive while the receiver is waiting for it, which is not necessarily what you want!

(Especially now while we don't yet have `.unload()`, I think NobodyWho Chat currently does weird tricks to make sure that `Chat` is deallocated when the application goes to the background, and this here might interfere with that).

Besides, I think we already handle the `None` case in `fn handle` below, so it shouldn't be a problem if the user does remove the chat before the receiver?

madsmtm: To be clear, I'd prefer we remove the guard here and the other `Arc` stuff you've added in this PR

nobodywho/core/src/chat.rs
```diff
@@ -1527,12 +1718,14 @@ fn process_worker_msg(worker_state: &mut Chat<'_>, msg: ChatMsg) {
...
-        ChatMsg::CompleteWithExternalTools {
+        ChatMsg::StructuredComplete {
             messages,
             options,
-            max_tokens,
+            mut max_tokens,
+            execute_tools,
             output_tx,
         } => {
+            max_tokens.get_or_insert(worker_state.engine.ctx.n_ctx() as usize);
```
madsmtm: Question: I think we're already limiting stuff to be within the context window elsewhere, is there a benefit to doing this again here?

nobodywho/core/src/sampler.rs
```diff
@@ -191,6 +191,48 @@ impl SamplerConfig {
...
+                self.sample_step = SampleStep::Dist;
+            }
+            replace_or_push_sampling_step(
+                &mut self.steps,
+                |step| matches!(step, ShiftStep::TopP { .. }),
+                ShiftStep::TopP { top_p, min_keep: 1 },
+            );
+        }
+        Ok(self)
+    }
```
madsmtm: Nit: I think this can be written more simply if you don't try to override an arbitrary sampler, but only do it as a construction method. Something like this:

```suggestion
    pub fn from_openrouter_parameters(
        temperature: Option<f32>,
        top_p: Option<f32>,
        seed: Option<u32>,
    ) -> Result<Self, SamplingOverrideError> {
        let temperature = temperature.unwrap_or(TEMPERATURE_DEFAULT);
        if !(0.0..=2.0).contains(&temperature) {
            return Err(SamplingOverrideError::InvalidTemperature);
        }

        let top_p = top_p.unwrap_or(TOP_P_DEFAULT);
        if !(0.0..=1.0).contains(&top_p) {
            return Err(SamplingOverrideError::InvalidTopP);
        }

        // TODO: Or random seed? What does OpenRouter do here?
        let seed = seed.unwrap_or_else(|| default_seed());

        if temperature == 0.0 {
            return Ok(SamplerPresets::greedy());
        }

        let mut builder = SamplerBuilder::new();

        builder = builder.shift(ShiftStep::default_top_k());
        builder = builder.shift(ShiftStep::TopP { top_p, min_keep: MIN_KEEP_DEFAULT });
        builder = builder.shift(ShiftStep::Temperature { temperature });

        builder = builder.seed(seed);

        Ok(builder.sample(SampleStep::Dist))
    }
```

madsmtm (changes requested):
Okay actually reviewed all of it this time, so should be good after resolving the comments (provided that I was detailed enough in them).

The only blockers are the guard Arc (https://github.com/nobodywho-ooo/nobodywho/pull/711#discussion_r4025681935) and that I'd prefer we only have the new wrappers (https://github.com/nobodywho-ooo/nobodywho/pull/711#discussion_r4026039316).

nobodywho/core/src/chat.rs
madsmtm: I'll leave this to your judgement

nobodywho/core/src/chat.rs
madsmtm: I'd prefer if we removed the old wrapper then, I don't see much benefit to keeping it?

E.g. merge `AssistantResponse`/`CompletionResponse` and merge `StructuredCompletionChunk`/`CompletionChunk`/`CompletionOutput` etc.?

duarteocarmo: Agreed - done.

nobodywho/core/src/chat.rs
duarteocarmo: Agreed - removed.

nobodywho/core/src/chat.rs
duarteocarmo: No. Removed.

nobodywho/core/src/sampler.rs
duarteocarmo: Nice! Re-wrote.

duarteocarmo:
@madsmtm  I think we should be covered now. Let me know if anything else comes up.

nobodywho/core/src/chat.rs
duarteocarmo: execute_tools now everywhere.

CHANGELOG.md
```diff
@@ -8,6 +8,14 @@ Format is based on [Keep a Changelog 1.1.0](https://keepachangelog.com/en/1.1.0/
...
 ## [Unreleased]
 
+### Added
+
+- **Rust:** Completion streams can report finish reasons and prompt and completion token counts. Per-request completion options now include output limits and OpenAI-style sampling parameters.
+
+### Changed
+
+- **Breaking, Rust:** `CompletionChunk::Done` now carries `CompletionResponse`, including usage and finish metadata. The duplicate metadata-specific response and stream types were removed.
+
```
madsmtm: Nit: We don't have a public Rust release yet, so I don't think these are necessary?

nobodywho/workspace.nix
```diff
@@ -90,6 +90,7 @@ let
         # `pyo3-ffi`". We bridge the names by appending an alias to the `env` file
         # that each crate installs and its dependents source.
         pyo3-ffi = attrs: {
+          nativeBuildInputs = (attrs.nativeBuildInputs or [ ]) ++ [ pkgs.python3 ];
```
madsmtm: Nit: I think this will be fixed by https://github.com/nobodywho-ooo/nobodywho/pull/745?

madsmtm (approved):
I like it, thanks!

PR #712, opened 2026-09-05
Compatible completions (3/3): Python APIs

duarteocarmo:
- Add the synchronous `NobodyWho` client with `chat.completions.create()` and `responses.create()`.
- Reuse core `Options` for sampling, keep output limits separate, and share one completion path for streaming and result metadata.
- Implement Responses as a small adapter that accepts `input` and turns `instructions` into a system message.
- Keep the existing `Chat` and `ChatAsync` APIs unchanged and document multi-turn and local tool calling.

### Chat completions multi-turn example

```python
from nobodywho import NobodyWho

client = NobodyWho()
model = "hf://NobodyWho/Qwen_Qwen3-0.6B-GGUF/Qwen_Qwen3-0.6B-Q4_K_M.gguf"
messages = [{"role": "user", "content": "My name is Duarte."}]

first = client.chat.completions.create(
    model=model,
    messages=messages,
    thinking=False,
)
messages.extend([
    {"role": "assistant", "content": first.choices[0].message.content},
    {"role": "user", "content": "What is my name?"},
])
second = client.chat.completions.create(
    model=model,
    messages=messages,
    thinking=False,
)
print(second.choices[0].message.content)
```

### Responses multi-turn example

```python
first = client.responses.create(
    model=model,
    instructions="Answer concisely.",
    input="My name is Duarte.",
    thinking=False,
)
second = client.responses.create(
    model=model,
    instructions="Answer concisely.",
    input=[
        {"role": "user", "content": "My name is Duarte."},
        *first.output,
        {"role": "user", "content": "What is my name?"},
    ],
    thinking=False,
)
print(second.output_text)
```

### Request flow

```mermaid
sequenceDiagram
    participant App as Python application
    participant Client as NobodyWho
    participant Responses as Responses adapter
    participant Completion as Completion path
    participant Cache as Model cache
    participant Core as Rust ChatHandle
    participant Worker as Inference worker

    alt Chat Completions
        App->>Client: chat.completions.create(messages, ...)
        Client->>Completion: Full message list and request options
    else Responses
        App->>Client: responses.create(input, instructions, ...)
        Client->>Responses: Input string or item list
        Responses->>Completion: Normalized messages and request options
    end
    Completion->>Cache: Resolve model
    Cache-->>Completion: Shared model
    Completion->>Core: Build chat and call complete_with_metadata(messages, options, max_tokens)
    Core->>Worker: Generate completion
    Worker-->>Completion: Token chunks and completion metadata
    alt Chat Completions
        Completion-->>Client: Chat completion chunks or result
    else Responses
        Completion-->>Responses: Chat completion stream or result
        Responses-->>Client: Response events or result
    end
    Client-->>App: Requested Python object
```


duarteocarmo:
Fixed the per-turn thinking persistence finding in 1482f2a0, using worker-owned request overrides from #711. Regression coverage checks both APIs, legacy and structured streams, errors, dropped streams, and persistent template-variable settings. Normal CI is running; full CI will follow once it passes.

duarteocarmo:
## Request flow

```mermaid
sequenceDiagram
    participant App as Python application
    participant Client as NobodyWho
    participant Cache as Model cache
    participant Binding as Python binding
    participant Core as Rust ChatHandle
    participant Worker as Inference worker

    App->>Client: chat.completions.create(...) or responses.create(...)
    Client->>Cache: Resolve model
    Cache-->>Client: Arc<Model>
    Client->>Binding: Convert messages and request options
    Binding->>Core: Build a fresh chat context
    Core->>Worker: complete_with_metadata(...)
    Worker-->>Binding: Token chunks
    Worker-->>Binding: CompletionResponse
    Binding-->>App: OpenAI-shaped result or stream
```

nobodywho/python/src/lib.rs
```diff
@@ -251,6 +309,396 @@ impl<'py> ModelOrPath<'py> {
...
+                });
+                Ok(model)
+            })
+            .map_err(|error| pyo3::exceptions::PyRuntimeError::new_err(error))?;
+        Ok((model, source))
+    }
+}
+
+struct RequestOptions {
+    sampler: Option<SamplerConfig>,
```
madsmtm, line 368: I think I would argue that we should get rid of this? It doesn't seem to be supported by OpenAI:
https://github.com/openai/openai-python/blob/d7c41efee1b0802b79f3f88a678ef2052b06e9ce/src/openai/resources/chat/completions/completions.py#L251-L298

And it complicates the logic (since you have to merge the temperature, top_p and seed into it).

nobodywho/python/src/lib.rs
```diff
@@ -27,6 +28,37 @@ fn py_completion_options(
...
 }
 
+fn structured_completion_options(
+    temperature: Option<f32>,
+    top_p: Option<f32>,
+    seed: Option<u32>,
+) -> PyResult<nobodywho::chat::Options> {
+    let sampler = nobodywho::sampler::SamplerConfig::default()
+        .with_sampling_overrides(temperature, top_p, seed)
+        .map_err(|error| pyo3::exceptions::PyValueError::new_err(error.to_string()))?;
```
madsmtm: I think I'd prefer for `with_sampling_overrides` to live here rather than in nw-core?

Or do you see a use-case for it outside of compatibility with the Python `openai` package?

(Perhaps I'm misunderstanding, is your plan to support other OpenRouter implementations on the other language bindings? Because if so, then I can better see the argument for keeping it in nw-core).

nobodywho/python/src/lib.rs
```diff
@@ -40,6 +72,7 @@ static PYTHON_LOGGING_AVAILABLE: AtomicBool = AtomicBool::new(false);
 #[pyclass]
 pub struct Model {
     model: Arc<nobodywho::llm::Model>,
+    source: String,
```
madsmtm: I think this would be useful to have directly in `nobodywho::llm::Model` and exposed everywhere?

Could you add this? (preferably in its own PR, either before or after this one, I think that would be fairly self-contained and easy to merge)

nobodywho/python/src/lib.rs
duarteocarmo: I think we will probably change some of the other APIs as well - that's why in core..

nobodywho/python/src/lib.rs
duarteocarmo: added a PR to the stack #743

nobodywho/python/src/bin/make_stubs.rs
```diff
@@ -16,6 +16,30 @@ const EXCEPTIONS_TO_REPLACE: &[(&str, &str)] = &[
...
+        "def __getitem__(self, /, key: str) -> builtins.object: ...",
+    ),
+    (
+        "def __getitem__(self, key: str, /) -> typing.Any: ...",
+        "def __getitem__(self, key: str, /) -> builtins.object: ...",
+    ),
+    (
+        "def __getitem__(self, key: str, /) -> Any: ...",
+        "def __getitem__(self, key: str, /) -> builtins.object: ...",
+    ),
```
madsmtm, line 42: I'm not sure I understand why we're doing this? Isn't `typing.Any` good enough (or actually more precise) here?

And if not, why are we only doing this here, and not replacing every `typing.Any` with `builtins.object` everywhere?

madsmtm: Ah, I see, we're doing this because we have a check later in the file that asserts that we don't use `typing.Any` anywhere...

docs/docs-python/chat.md
```diff
@@ -9,6 +9,124 @@ In the following sections, we talk about which configuration options it has, and
 
 You can pass `"auto"` as the model path to select a chat model based on available memory.
 
+## OpenAI API compatibility
+
+`NobodyWho` provides local Python APIs that follow the main request and response shapes from OpenAI's Chat Completions and Responses APIs. You use `client.chat.completions.create()` or `client.responses.create()` with a local model. These APIs implement compatible subsets of the OpenAI APIs. They are not complete replacements for the OpenAI Python SDK.
+
+Both APIs are stateless. Pass the full conversation on every call. The client caches the last model source, but each request gets a new context and new settings.
```
madsmtm: Nit:

```suggestion
Both APIs are stateless, so you need to pass the full conversation on every call. The client caches the model and KV cache etc. when possible.
```

docs/docs-python/chat.md
```diff
@@ -9,6 +9,124 @@ In the following sections, we talk about which configuration options it has, and
...
+   messages=[{"role": "user", "content": "What is the weather in Copenhagen?"}],
+   tools=[get_weather],
+   thinking=False,
+)
+print(completion.choices[0].message.content)
+```
+
+NobodyWho does not implement every OpenAI option or response field. Missing features include multiple choices, structured output through `response_format`, log probabilities, penalties, stop sequences, audio output, OpenAI hosted tools, and OpenAI storage and service metadata. Streaming returns local Python objects rather than OpenAI SDK stream objects.
+
+See the [OpenAI Chat Completions API reference](https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create).
```
madsmtm, line 73: Nit: Maybe link to the API we're mimicking more at the start instead of here? E.g.:
> NobodyWho supports [the main Chat Completions flow](...) with

And similarly for the Responses API:
> NobodyWho supports [the main Responses text flow](...)

nobodywho/python/src/lib.rs
```diff
@@ -27,6 +28,37 @@ fn py_completion_options(
     }
 }
 
+fn compatible_completion_options(
```
madsmtm, line 31: Nit: I'd prefer to inline this into `request_completion_stream`, it's small enough (now) that I think it makes sense that way instead.

nobodywho/python/src/lib.rs
```diff
@@ -334,6 +378,383 @@ impl<'py> ModelOrPath<'py> {
...
+        .with_context_size(n_ctx)
+        .with_template_variables(template_variables)
+        .with_tools(tools.into_iter().map(|tool| tool.tool).collect());
+    if let Some(n_threads) = n_threads {
+        builder = builder.with_n_threads(n_threads);
+    }
+    if let Some(mtp) = mtp {
+        builder = builder.with_mtp(mtp.into());
+    }
+    let chat = py.detach(|| builder.build()).map_err(err)?;
```
madsmtm, line 478: Oof, I'd greatly prefer if we didn't create a new `Chat` object every call, but instead only do so when something actually changed - prefilling is expensive and if we can avoid it we should?

Maybe at least leave a FIXME or TODO to improve the performance here (it might need work in `core` as well to do things properly, so I'd be okay with doing it later).

nobodywho/python/src/lib.rs
```diff
@@ -948,6 +1369,475 @@ impl TokenStream {
...
+    parser: ReasoningStreamParser,
+    pending_deltas: VecDeque<(ParsedResponseText, String)>,
+    pending_response: Option<nobodywho::chat::CompletionResponse>,
+    completed: Option<ChatCompletion>,
+}
+
+impl Drop for ChatCompletionStream {
+    fn drop(&mut self) {
+        let inner = self.inner.take();
+        Python::attach(|py| py.detach(|| drop(inner)));
```
madsmtm, line 1456: Huh? Isn't this a no-op? Pretty sure holding the GIL is a reference-counted-like state, so this doesn't allow `inner`'s drop glue to run outside the GIL.

madsmtm: Oh, I see we're doing the same thing elsewhere... Which is probably wrong, but fine to keep the same pattern here.

nobodywho/python/src/lib.rs
```diff
@@ -334,6 +378,383 @@ impl<'py> ModelOrPath<'py> {
...
+    let (model, model_source) = state.model_for(request.model, py)?;
+    let inner = request_completion_stream(model, request.messages, request.options, py)?;
+    Ok(ChatCompletionStream::new(
+        inner,
+        model_source,
+        request.id_prefix,
+    ))
+}
+
+static COMPLETION_IDS: AtomicU64 = AtomicU64::new(1);
```
madsmtm, line 498: Q: Are you sure these are global and not scoped to the session?

nobodywho/python/src/bin/make_stubs.rs
madsmtm: Although, does `#[pyo3(signature = ...)]` not work in dunder-methods, or why aren't we just using that? That seems less error-prone?

nobodywho/python/src/lib.rs
```diff
@@ -334,6 +378,383 @@ impl<'py> ModelOrPath<'py> {
...
+    serde_json::to_value(value).map_err(|error| {
+        pyo3::exceptions::PyRuntimeError::new_err(format!("failed to serialize response: {error}"))
+    })
+}
+
+fn response_field(value: &impl Serialize, key: &str, py: Python<'_>) -> PyResult<Py<PyAny>> {
+    let value = serialized_response(value)?;
+    let field = value
+        .get(key)
+        .ok_or_else(|| pyo3::exceptions::PyKeyError::new_err(key.to_owned()))?;
```
madsmtm, line 510: Remark: There's probably more performant ways of doing this, but it's simple and kinda elegant in a way, so I'm fine with it.

nobodywho/python/src/lib.rs
```diff
@@ -334,6 +378,383 @@ impl<'py> ModelOrPath<'py> {
...
+response_mapping_methods!(ChatCompletionMessage);
+response_mapping_methods!(ChatCompletionChoice);
+response_mapping_methods!(ChatCompletion);
+response_mapping_methods!(ChatCompletionDelta);
+response_mapping_methods!(ChatCompletionChunkChoice);
+response_mapping_methods!(ChatCompletionChunk);
+response_mapping_methods!(ResponseOutputText);
+response_mapping_methods!(ResponseOutputMessage);
+response_mapping_methods!(ResponseUsage);
+response_mapping_methods!(ResponseStreamEvent);
```
madsmtm, line 717: Nit: I think I'd prefer to have these right under their respective struct, including moving the `impl Response` up under `struct Response`, that makes it easier to see that `Response` doesn't have `response_mapping_methods!`?

nobodywho/python/src/lib.rs
```diff
@@ -948,6 +1369,475 @@ impl TokenStream {
...
+                self.reasoning = !self.reasoning;
+                continue;
+            }
+            let held = if finished {
+                0
+            } else {
+                (1..marker.len())
+                    .rev()
+                    .find(|length| self.pending.ends_with(&marker[..*length]))
+                    .unwrap_or(0)
```
madsmtm, line 1414: Remark: Took me a while to figure out that this is buffering stuff like `<thi` if that is in the output, but emitting everything else. Would have been useful to have a comment explaining this.

madsmtm: (Relatedly, I think that parsing token strings is probably a bad idea, we should parse based on the actual LlamaToken, not whatever its textual representation happens to be. Which is yet another argument for doing this properly in core, but let's defer)

nobodywho/python/src/lib.rs
```diff
@@ -948,6 +1369,475 @@ impl TokenStream {
...
+        parsed.push((
+            if self.reasoning {
+                ParsedResponseText::Reasoning
+            } else {
+                ParsedResponseText::Content
+            },
+            text,
+        ));
+    }
+}
```
madsmtm, line 1437: I suspect this would be possible to simplify if `core` parsed tokens more / knew what state the token stream was in at any given time, cc @jona605a.

nobodywho/python/tests/test_nobodywho.py
```diff
@@ -26,6 +26,11 @@ def chat(model):
     )
```
madsmtm, line 26: Q: Do you think it would make sense to grab 3 [random projects using the `openai` package](https://www.wheelodex.org/projects/openai/rdepends/) (or [here](https://github.com/openai/openai-python/network/dependents?package_id=UGFja2FnZS01MjM3OTYwMw%3D%3D)), and evaluating how easy that is / what works, what doesn't?

nobodywho/python/src/lib.rs
```diff
@@ -3585,6 +4873,62 @@ fn get_cached_models() -> PyResult<Vec<(String, usize)>> {
         .collect()
 }
 
+#[cfg(test)]
+mod completion_tests {
```
madsmtm, line 4877: Nit: Add the module at the very bottom of the file, below `nobodywhopython`, and call it just `tests`

nobodywho/python/src/lib.rs
```diff
@@ -2316,8 +3206,406 @@ impl ChatAsync {
...
+    }
+
+    if !system_parts.is_empty() {
+        converted.insert(
+            0,
+            nobodywho::chat::Message::new_system(system_parts.join("\n\n")),
+        );
+    }
+    Ok(converted)
+}
```
madsmtm, line 3458: Remark: I haven't really reviewed this, my eyes are glazing over it. Feels like more of it should be able to use `serde` though? At least the `role` could probably be tagged enums.

That would make it a lot easier for me to parse.

madsmtm (approved):
Not the review I'm the most proud of, partially because it took me so long to get to, but the code is mostly about shuffling data from one format to another, so it should be relatively low-risk. The parts I'm the most weary about are the parts with logic in them, like `next_chunk` and `ReasoningStreamParser`, dunno of a way to assuage that apart from adding more tests for them, or eventually supporting them more formally in `core`?

Gave a bunch of not really important nits and remarks on my way, feel free to ignore them.

I'll note that I haven't compared this much to the OpenAI APIs that we're mimicking, so I am not really able to tell how well this actually fulfills its purpose? See https://github.com/nobodywho-ooo/nobodywho/pull/712#discussion_r4040769255 for discussion on one way to improve that.

nobodywho/python/tests/test_nobodywho.py
madsmtm: Or maybe simpler and better, would it be possible for us to run some of [the `openai` examples](https://github.com/openai/openai-python/tree/main/examples) with minimal modification?

nobodywho/python/src/lib.rs
jona605a: Yeah, that is one of the key outcomes of a better token stream! Core will handle that, so the other end of the stream doesn't have to

PR #715, opened 2026-09-07
Expose grammar constraints in samplerbuilder

jona605a:
Fixes issue #693 and linear issue 140 (essentially the same).

Overall pretty simple change, then some minor details.

`SamplerBuilder` now exposes `constrain_with_json_schema`, `constrain_with_regex`, `constrain_with_grammar` and `json`, so these can be chained with other user-specified sampler options, instead of just a plain `SamplerPresets::json` where you can't change the `top_k` or `top_p`, for example.

I also changed the `json()` sampling from using the hand-written Lark grammar to using `ShiftStep::JsonSchema(r#"{"type":"object"}"#)`. This removes the `const JSON_GRAMMAR` but also fixes a behavioral bug where the old grammar would accept an empty object and a model (I tested Qwen3-0.6B) could answer `{}` to "give me a person as JSON with a name field".
This also gets the json slices from `JsonSchema` - slightly more precompute for faster parsing.

I deleted `ShiftStep::LarkWithSlices` as it wasn't used anywhere in the code. Because `ShiftStep` is `#[serde(tag = "type")]` and `SamplerConfig.from_json()` is public API in the bindings, it was technically reachable, so I added a changelog note about it.

PR #716, opened 2026-09-08
SamplerPresets use SamplerConfig::Defaults plus their own change

jona605a:
Builds on #715 . 

Instead of `SamplerPresets` like `temperature` only setting the temperature and not having any `top_k` or `top_p` values, giving potentially bad output since unlikely tokens can be generated, this PR makes all `SamplerPresets` use the `SamplerConfig::Default` and just add/change one option.
Exception is `greedy`, since none of the default steps change its choice.


nobodywho/uniffi/src/lib.rs
madsmtm, line 2014: Nit: I know it's pre-existing, but could you add the same docs here as elsewhere?
> Constrain output to a JSON object of any shape.

docs/docs-python/sampling.md
```diff
@@ -35,6 +35,10 @@ class SamplerPresets:
     def constrain_with_grammar(grammar: str) -> SamplerConfig: ...
 ```
 
+Every preset starts from `default()` and changes one thing, so `constrain_with_json_schema` still
+samples with the default top-k, top-p and temperature. `greedy()` is the exception: it always picks
+the most probable token, so it needs no other steps.
```
madsmtm: Nit: Reads a bit weirdly IMO, maybe something like this would be better?

```suggestion
All presets have top-k, top-p, temperature and dist steps (except for `greedy()`
which always picks the most probable token, so it needs no other steps).
```

docs/docs-python/sampling.md
madsmtm: Remark: I tend to think this is a bad idea? We should instead raise an exception if you add the steps in the wrong order?

nobodywho/core/src/sampler.rs
```diff
@@ -13,23 +13,54 @@ use crate::errors::SamplerError;
...
+/// top, replacing the default step of the same kind if there is one. So
+/// `constrain_with_json_schema` still samples with the default top-k, top-p and
+/// temperature, and `temperature` overrides only the temperature. The exception
+/// is [`SamplerPresets::greedy`], which always picks the most probable token and
+/// so needs no shift steps at all.
 pub struct SamplerPresets;
 
+/// The default steps, with `replacement` substituted for the default step of the
+/// same kind (appended if the defaults have no such step).
+fn defaults_replacing(replacement: ShiftStep) -> Vec<ShiftStep> {
```
madsmtm: I know I know, DRY etc., but I feel like this is somewhat over-complicating it?

I'd prefer that each preset just spelled out exactly what steps it has:
```rust
pub fn top_k(k: i32) -> SamplerConfig {
    SamplerConfig::new(
        vec![
            ShiftStep::TopK { top_k: k },
            ShiftStep::TopP {
                top_p: 0.95,
                min_keep: 1,
            },
            ShiftStep::Temperature { temperature: 0.6 },
        ],
        SampleStep::Dist,
        default_seed(),
    )
}
```

If you want to reduce duplication, then I'd rather have functions like `ShiftStep::default_top_k()`, `ShiftStep::default_top_p()` and `ShiftStep::default_temperature()`.

WDYT?

nobodywho/core/src/sampler.rs
```diff
@@ -412,17 +450,19 @@ impl SamplerBuilder {
...
-            | ShiftStep::Lark(_) => self.steps.insert(0, step),
-            _ => self.steps.push(step),
+        if step.runs_before_truncation() {
+            // The chain is always a run-first prefix followed by the rest.
+            let at = self
+                .steps
+                .partition_point(ShiftStep::runs_before_truncation);
+            self.steps.insert(at, step);
+        } else {
+            self.steps.push(step);
```
madsmtm: Remark: I tend to think that reordering the steps is kind of a bad idea? It feels surprising to me that the user doesn't actually control in which order the samplers go.

Maybe we should instead raise an exception if you add the steps in the wrong order?

Or expand on the existing `ShiftStep`/`SampleStep` split into something like:
```rust
pub struct SamplerConfig {
    pub constraining_steps: Vec<ConstraintStep>,
    pub steps: Vec<ShiftStep>,
    pub sample_step: SampleStep,
    pub seed: u32,
}
```

CHANGELOG.md
```diff
@@ -12,13 +12,14 @@ Format is based on [Keep a Changelog 1.1.0](https://keepachangelog.com/en/1.1.0/
...
-- `SamplerPresets.json()` and the deprecated `SamplerPresets.grammar()` no longer prepend their constraint to the default sampling chain (top-k 20, top-p 0.95, temperature 0.6) — they constrain and sample directly, as `constrain_with_json_schema`/`_regex`/`_grammar` already did. Output is still always valid, but picks within the constrained set are more random than before. Affects all bindings.
+- Every `SamplerPresets` entry now builds on the default sampler (top-k 20, top-p 0.95, temperature 0.6) and changes one thing, instead of producing a chain holding only its own step: enabling a constraint no longer drops the truncation and temperature you would otherwise be sampling with, and `top_k`, `top_p` and `temperature` each override their counterpart and leave the rest alone. Constrained output is valid as before but less random within the constrained set. `greedy` is unaffected — it needs no shift steps. Affects all bindings.
```
madsmtm: This is a breaking change, and should perhaps be noted more visibly?

madsmtm (changes requested):
I think this is a great idea, sampler presets are a lot more useful with this!

nobodywho/core/src/sampler.rs
jona605a: Hmm, then I think maybe we could have something like
```rust
const TOP_K_DEFAULT: i32 = 20;
const TOP_P_DEFAULT: f32 = 0.95;
const MIN_KEEP_DEFAULT: i32 = 1;
const TEMPERATURE_DEFAULT: f32 = 0.6;

pub fn top_k(k: i32) -> SamplerConfig {
    SamplerConfig::new(
          vec![
              ShiftStep::TopK { top_k: k },
              ShiftStep::TopP {
                  top_p: TOP_P_DEFAULT,
                  min_keep: MIN_KEEP_DEFAULT,
              },
              ShiftStep::Temperature { temperature: TEMPERATURE_DEFAULT },
          ],
          SampleStep::Dist,
          default_seed(),
        )
}
```

nobodywho/core/src/sampler.rs
jona605a: Oh, I like that suggestion a lot

jona605a:
Slight scope creep from initial PR description:
I took Mads' suggestion, and refactored `sampler.rs` so that `SamplerConfig::constraining_steps` is now the place where constraining steps live instead of among the other shift steps. 

I also took the liberty to remove the `grammar()` preset along with its `build_lazy_grammar` and `build_regular_grammar` functions.

nobodywho/core/src/sampler.rs
madsmtm, line 173: Let me verify I understand things right: At a high level, what we want to do when sampling is to take a list of logits `[0.4, 0.2, 0.5, ...]` and select a single index (token) from that list?

My understanding then is that there's basically three intermediate operations you can perform on the list of logits:
- Filter (remove logits from consideration like `[0.4, 0.2, 0.5, ...] -> [0.4, 0.5, ...]`)
- Map (change the value of logits like `[0.4, 0.5, ...] -> [0.2, 0.8, ...]`)
- Fold/reduce (select a value from the remaining list of logits `[0.2, 0.8, ...] -> index 1`)

Does that make sense? Is filtering analogous to `ConstrainStep` and mapping analogous to `ShiftStep`, or am I misunderstanding? If not, under this more PL-like terminology, which samplers would fit where?

nobodywho/core/src/chat.rs
```diff
@@ -3967,7 +3967,7 @@ mod tests {
             .build()
             .expect("chat build failed in test");
 
-        let greedy = SamplerConfig::new(vec![], crate::sampler::SampleStep::Greedy, 1234);
+        let greedy = SamplerConfig::new(vec![], vec![], crate::sampler::SampleStep::Greedy, 1234);
```
madsmtm: Nit: Maybe?

```suggestion
        let greedy = SamplerPresents::greedy();
```

docs/docs-python/sampling.md
```diff
@@ -160,7 +160,11 @@ Shift steps — add as many as you want, applied in order:
 - `.logit_bias({ 1: -1.0, 2: 3.0 })` — token 1 less probable, token 2 is more probable
 - `.dry(0.8, 1.75, 2, -1, ["\n"])` — penalty for repeated *phrases*: `multiplier, base, allowed_length, last_n, seq_breakers`
 - `.seed(42)` — fix the RNG for reproducible output
-- `.grammar(...)` — deprecated; use `.constrain_with_grammar(...)` below
+
+The order you chain them matters: `.penalties(...)`, `.logit_bias(...)` and `.dry(...)`
+reweight whatever distribution reaches them, so put them *before* any truncation step if you
```
madsmtm: Nit: English?

```suggestion
reweigh whatever distribution reaches them, so put them *before* any truncation step if you
```

madsmtm: Q: What does "truncation step" mean in this sentence?

docs/docs-python/sampling.md
```diff
@@ -160,7 +160,11 @@ Shift steps — add as many as you want, applied in order:
...
+want them to see the whole vocabulary — that is why llama.cpp's own default chain leads with
+the penalties.
```
madsmtm: So why doesn't _our_ default chains lead with penalties?

CHANGELOG.md
```diff
@@ -12,13 +12,15 @@ Format is based on [Keep a Changelog 1.1.0](https://keepachangelog.com/en/1.1.0/
...
+- **Breaking:** every `SamplerPresets` entry now builds on the default sampler (top-k 20, top-p 0.95, temperature 0.6) and changes one thing, instead of producing a chain holding only its own step: enabling a constraint no longer drops the truncation and temperature you would otherwise be sampling with, and `top_k`, `top_p` and `temperature` each override their counterpart and leave the rest alone. Constrained output is valid as before but less random within the constrained set. `greedy` is unaffected — it needs no shift steps. Affects all bindings.
+- `SamplerPresets.dry()` now actually applies the DRY penalty. Its multiplier was 0.0, which llama.cpp reads as "disabled", so the preset was a no-op that sampled exactly like the default one. It is now 0.8, the value the preset's other numbers (base 1.75, allowed length 2) are tuned for. The preset leads with its DRY step, so the penalty sees the whole vocabulary rather than what survived truncation. Affects all bindings.
```
madsmtm: Remark: I'm not gonna bother reading this, whatevs.

madsmtm (approved):
I like it, I'll leave it up to you as to whether and how we should rename `ConstraintStep` to `GrammarStep`

docs/docs-python/sampling.md
jona605a: grammar step. Can re word it

docs/docs-python/sampling.md
jona605a: That wording was confusing - I've changed it in the commit below. Penalties are repetition-penalties, i.e. they make the model generate more varied text. In case it _is_ enabled, then you want to have it in the beginning of the chain because you don't want to include some tokens in your `top_k(10)` tokens only for half of them to be penalized. What llama.cpp does is reserve a *spot* for the penalty but their default sampler doesn't use it. It's just there so it can be configured/enabled.

PR #719, opened 2026-09-09
:sparkles: Llama cpp style model names

albertsgarde:
Add support for the model format that [llama.app](llama.app) uses.
For example you can now do
```
from nobodywho import Model

model = Model("ggml-org/Qwen3.8-27B-GGUF:Q8_0")
```
and you get the `Qwen3.8-27B` model with `Q8_0` quantization.

Unlike Llama cpp, we do not support no quantization. So only writing `ggml-org/Qwen3.8-27B-GGUF` would work with Llama cpp, but not with us. There are two reasons for this. Firstly, it would be ambiguous whether the user meant a file path or a huggingface model. Secondly, the way Llama cpp handles these unspecified quantizations is by choosing the first quantization found in the HF repo. To do that we would need to request HF during URL resolution which I would rather avoid. It also gives a much simpler implementation.

I also added Rust tests for both basic parsing of the format (and for the HuggingFace format which wasn't being tested before) as well as a test for whether the URL of the model file is resolved correctly, with test cases taken directly from the models `llama.app` supports.


krebso:
Lets address the P2, otherwise LGTM & good job

albertsgarde:
@krebso I've fixed the things you mentioned. Approve pls 🥺?

krebso (approved):

PR #721, opened 2026-09-09
Make `InferenceEngine` return one token at a time

madsmtm:
Refactor `InferenceEngine` to store the draft data internally, such that it can return one token at a time. Ideally, it should be opaque to the caller that `InferenceEngine::next_token` can choose to generate multiple tokens when using speculative decoding / MTP.

This makes it easier for `Chat` to consume, since it doesn't need to worry about context overflow etc. inside the inference engine.

This should also make it possible in the future to refactor `Chat` to work on a more pull-based model, where the caller "pulls" tokens instead of the `Chat` running in a background thread and eagerly pushing them onto a queue.

### DIFF
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -1641,13 +1641,7 @@ fn process_worker_msg(worker_state: &mut Chat<'_>, msg: ChatMsg) {
             let _ = output_tx.blocking_send(stats);
         }
         ChatMsg::GetMtpAcceptanceRate { output_tx } => {
-            let proposed = worker_state.engine.mtp_drafts_proposed;
-            let rate = if proposed > 0 {
-                Some(worker_state.engine.mtp_drafts_accepted as f32 / proposed as f32)
-            } else {
-                None
-            };
-            let _ = output_tx.blocking_send(rate);
+            let _ = output_tx.blocking_send(worker_state.engine.mtp_acceptance_rate());
         }
         ChatMsg::Tokenize { prompt, output_tx } => {
             let result = worker_state.tokenize(prompt);
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -1636,7 +1636,7 @@ fn process_worker_msg(worker_state: &mut Chat<'_>, msg: ChatMsg) {
         ChatMsg::GetStats { output_tx } => {
             let stats = ChatStats {
                 context_size: worker_state.engine.ctx.n_ctx(),
-                context_used: worker_state.engine.n_past(),
+                context_used: worker_state.engine.actual_context_size() as u32,
             };
             let _ = output_tx.blocking_send(stats);
         }
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -2777,7 +2777,8 @@ mod tests {
     /// env vars are set to existing files.
     #[test]
     fn test_mtp_gemma4_smoke() -> Result<(), Box<dyn std::error::Error>> {
-        // test_utils::init_test_tracing();
+        test_utils::init_test_tracing();
+
         let (Some(target_path), Some(draft_path)) = (
             test_utils::test_mtp_target_model_path(),
             test_utils::test_mtp_draft_model_path(),
@@ -2822,11 +2823,21 @@ mod tests {
                 sender.send(resp).unwrap();
             }
         };
-
         worker.ask("What is the capital of Denmark?".into(), f)?;
         let resp = receiver.recv()?;
         println!("MTP response: {}", resp);
         assert!(resp.contains("Copenhagen"));
+
+        let (sender, receiver) = std::sync::mpsc::channel();
+        let f = move |x| {
+            if let llm::WriteOutput::Done(resp) = x {
+                sender.send(resp).unwrap();
+            }
+        };
+        worker.ask("Are you sure?".into(), f)?;
+        let resp = receiver.recv()?;
+        println!("MTP response: {}", resp);
+        assert!(resp.contains("Yes"));
 
         Ok(())
     }
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -2089,7 +2089,6 @@ impl<'a> Chat<'a> {
         // 4096 is a very randomly chosen number. how does this affect performance?
         let mut full_response: String = String::with_capacity(4096);
         let mut tokens_written_until_now = Vec::new();
-        let mut new_tokens = Vec::new();
         let mut generated_tokens = 0;
 
         self.sampler.reset();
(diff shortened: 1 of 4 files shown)


nobodywho/core/src/inference.rs
```diff
@@ -327,25 +325,27 @@ impl<'a> InferenceEngine<'a> {
...
+        // Keep the MTP draft ctx's hidden state in sync.
+        if let EngineContext::Speculative(spec) = &mut self.ctx {
+            spec.process(&self.batch)?;
+        }
 
         self.n_past += tokens.len() as i32;
-        // A new prompt (or context-shift replay) invalidates any deferred
-        // pending sample from a previous generation.
-        self.pending = None;
+        // A new prompt (or context-shift replay) invalidates any drafts
```
madsmtm: I feel like there might be bugs here in us not calling `spec.accept`? But unsure whether we should "accept" newly added tokens that weren't drafted?

nobodywho/core/src/inference.rs
```diff
@@ -383,11 +383,52 @@ impl<'a> InferenceEngine<'a> {
...
+            let EngineContext::Speculative(spec) = &mut self.ctx else {
+                unreachable!("only context should not have drafts");
+            };
+            spec.accept(self.draft_state.accepted as u16)?;
+        }
+
+        self.mtp_drafts_accepted += self.draft_state.accepted as u64;
+
+        Ok(())
+    }
```
madsmtm: I am not suuuper happy about how we handle this state, but I think it's necessary to support the way that MTP works in llama.cpp, where it requires a `spec.accept(accepted)` call (a bit unsure why?)

nobodywho/core/src/inference.rs
```diff
@@ -457,137 +475,58 @@ impl<'a> InferenceEngine<'a> {
             .load_audio(path)
     }
 
-    pub(crate) fn sample_and_decode_next_tokens(
+    /// Sample and decode the next token.
+    pub(crate) fn next_token(
```
madsmtm, line 541: This used to be two methods (one for solo, one for mtp), but I ended up merging it because I found it kinda nice that you can see the similarities between the two (in particular that we sample first, and then decode).

nobodywho/core/src/chat.rs
```diff
@@ -2113,77 +2110,54 @@ impl<'a> Chat<'a> {
...
-            self.engine.sample_and_decode_next_tokens(
-                &mut self.sampler,
-                &mut new_tokens,
-                remaining_tokens,
-            )?;
-
-            tokens_written_until_now.extend_from_slice(&new_tokens);
-
-            let mut should_finish = false;
-            for &new_token in &new_tokens {
```
madsmtm, line 2134: This inner inner loop is now gone, that was the point!

nobodywho/core/src/inference.rs
```diff
@@ -457,171 +535,106 @@ impl<'a> InferenceEngine<'a> {
...
+        //
+        // The second point in particular is important for performance:
+        // ideally, we always want a decoding stage in progress, so that all
+        // the various other work we do (including the work the user does)
+        // isn't going to block inference.
+
+        let span = trace_span!("sample").entered();
+        let token = if let EngineContext::Speculative(spec) = &mut self.ctx {
+            if let Some(draft) = spec.drafts.get(spec.accepted) {
+                let token = sampler.sample(spec.ctx.target_context_mut(), spec.accepted as _);
```
jona605a, line 564: `as _` - wow you can do that? Cool

nobodywho/core/src/inference.rs
```diff
@@ -52,6 +52,76 @@ where
...
+
+        self.total_proposed += self.drafts.len() as u64;
+        self.total_accepted += self.accepted as u64;
+
+        Ok(())
+    }
+
+    fn roll_back_declined_drafts(&mut self, keep_up_to: u32) -> Result<(), RollbackError> {
+        let declined = self.drafts.len() - self.accepted;
+        if 0 < declined {
```
jona605a, line 103: This inequality order is cursed :P

nobodywho/core/src/inference.rs
```diff
@@ -52,6 +52,76 @@ where
...
 }
 
+/// MTP state.
+#[derive(Debug)]
+pub(crate) struct SpeculativeEngine<'a> {
+    ctx: MtpSpeculative<'a>,
+    /// The in-progress drafts.
+    drafts: Vec<LlamaToken>,
+    /// The number of accepted drafts.
+    accepted: usize,
```
jona605a, line 62: Nit: maybe call it `n_accepted` to indicate it's a count and not a list / object?

nobodywho/core/src/inference.rs
```diff
@@ -457,171 +535,106 @@ impl<'a> InferenceEngine<'a> {
...
-            output.push(pending);
-            return Ok(());
+        self.batch.clear();
+        self.batch.add(token, self.n_past, &[0], true)?;
+        for (i, &d) in drafts.iter().enumerate() {
+            self.batch.add(d, self.n_past + 1 + i as i32, &[0], true)?;
         }
 
-        sampler.observe(pending);
+        let span = trace_span!("decode", n_past = self.n_past).entered();
```
jona605a, line 620:
```suggestion
        // llm go brr
        let span = trace_span!("decode", n_past = self.n_past).entered();
```

It's important to keep our comments consistent between this and `InferenceEngine::read_text_tokens` ;)
(and to somehow communicate that this is where a lot of compute happens)

jona605a (approved):
Very nice re-write! Would be nice if the update / management of the `SpeculativeEngine` state (like `accept_drafts()`) were a bit more explained with a comment.

nobodywho/core/src/inference.rs
madsmtm: Yeah, the compiler can easily see that `.sample` wants an `i32`.

It's kind of a hack in any case though, `as` casts have a bunch of pitfalls, see this Clippy lint for details:
https://rust-lang.github.io/rust-clippy/master/index.html#as_conversions

It's just sooo much more convenient, and realistically, we're not going to have more than `i32::MAX` accepted drafts ;).

nobodywho/core/src/inference.rs
madsmtm: You think? I think I'm biased from reading https://matklad.github.io/2023/09/13/comparative-analysis.html

nobodywho/core/src/inference.rs
madsmtm: The spans help communicate that: When you run with `.with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE)`, as we do in tests, you get timing details like:
```
$ cargo test -- test_reset_chat --no-capture
...
   0.700309792s TRACE next_token:sample: nobodywho::inference: close time.busy=4.82ms time.idle=3.42µs
   0.701002625s TRACE next_token:decode{n_past=35}: nobodywho::inference: close time.busy=663µs time.idle=2.62µs
   0.701021000s TRACE next_token: nobodywho::inference: close time.busy=5.54ms time.idle=2.00µs
   0.701039667s TRACE nobodywho::chat: new_token=LlamaToken(32313) token_str="Okay" has_eog=false
   0.701048875s TRACE nobodywho::chat: Sending out token: token_str="Okay"
...
```

More importantly, this tells you that, in fact, **llm no go brr** here! Sampling took 4.8ms, while decoding only to 0.7ms!

(This is because `decode` spawns tasks that are executed asynchronously, and sampling waits for those to finish, as I noted in the comment at the top of `InferenceEngine::next_token`. Does that make sense?)

nobodywho/core/src/inference.rs
madsmtm: Added a FIXME

nobodywho/core/src/inference.rs
madsmtm: Added a FIXME

madsmtm:
> Very nice re-write! Would be nice if the update / management of the `SpeculativeEngine` state (like `accept_drafts()`) were a bit more explained with a comment.

I tried to write something in the last commit, let me know what you think?

PR #723, opened 2026-09-09
Release: :bookmark: Changelog, docs, version bumps

albertsgarde:
Bump all versions, regenerate docs, and update changelog.


albertsgarde:
Issues with the changelog are known, but no possible to fix properly before this is merged. I have a task to fix the changelog with a PR afterwards

s194042 (approved):

PR #726, opened 2026-09-10
fix: commit assistant message before tool call correctly to history

jona605a:
bug fix: when the model generated a message with a tool call, we streamed the entire response but didn't commit the pre-tool call content to `self.messages`. Now added.

### DIFF
--- a/nobodywho/core/src/chat.rs
+++ b/nobodywho/core/src/chat.rs
@@ -58,8 +58,8 @@ pub enum Message {
         content: MessageContent,
     },
     // The optional tool_calls field distinguishes a plain assistant response
-    // from one that includes tool calls. When tool_calls is Some, the content
-    // field is typically empty (required by qwen3 chat templates).
+    // from one that includes tool calls. When tool_calls is Some, content holds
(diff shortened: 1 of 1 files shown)


nobodywho/core/src/chat.rs
```diff
@@ -58,8 +58,8 @@ pub enum Message {
         content: MessageContent,
     },
     // The optional tool_calls field distinguishes a plain assistant response
-    // from one that includes tool calls. When tool_calls is Some, the content
-    // field is typically empty (required by qwen3 chat templates).
+    // from one that includes tool calls. When tool_calls is Some, content holds
+    // whatever the model said before the first call.
     // https://github.com/QwenLM/Qwen3/blob/e5a1d326/docs/source/framework/function_call.md
```
jona605a, line 63: This previous comment is straight up wrong. The link says "The template in the tokenizer_config.json assumes that the generated content alongside tool calls is in the same message instead of separate assistant messages".

duarteocarmo (approved):
Nice

PR #729, opened 2026-09-10
Handle Python download interruptions

duarteocarmo:
- Cancel synchronous Python GGUF downloads when Ctrl+C is pressed and preserve `KeyboardInterrupt`.
- Remove incomplete temporary files and cover `Model`, `download_model`, and model-path consumers.
- Add local HTTP regression tests for cancellation and cleanup.
- Add the Python changelog entry and changelog guidance for agents.
- Checks: Rust formatting and compile checks, plus Ruff formatting and linting.

### DIFF
--- a/nobodywho/core/src/huggingface.rs
+++ b/nobodywho/core/src/huggingface.rs
@@ -29,6 +29,18 @@ use tracing::{info, warn};
 /// synchronization (hence the `Sync` bound).
 pub type DownloadProgressCallback = Arc<dyn Fn(u64, u64) + Send + Sync>;
 
+/// Callback checked between download chunks. Returning `true` stops the download.
+pub type DownloadCancellationCallback = Arc<dyn Fn() -> bool + Send + Sync>;
+
+fn check_download_cancellation(
+    cancellation: Option<&DownloadCancellationCallback>,
+) -> Result<(), LoadModelError> {
+    if cancellation.is_some_and(|is_cancelled| is_cancelled()) {
+        return Err(LoadModelError::DownloadCancelled);
+    }
+    Ok(())
+}
+
 /// Default terminal progress bar shown when the user doesn't pass their own callback,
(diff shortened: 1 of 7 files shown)


nobodywho/core/src/huggingface.rs
```diff
@@ -307,12 +322,14 @@ impl ModelCache {
...
         content_length: u64,
         progress: &DownloadProgressCallback,
+        cancellation: Option<&DownloadCancellationCallback>,
     ) -> Result<(), LoadModelError> {
         let mut downloaded: u64 = 0;
         let mut last_logged_pct: u64 = 0;
         let mut buf = vec![0u8; 256 * 1024];
 
         loop {
+            check_download_cancellation(cancellation)?;
```
jona605a, line 332: This is checked twice per loop? Is this one needed?

nobodywho/python/src/lib.rs
```diff
@@ -41,22 +42,91 @@ pub struct Model {
...
 }
 
-/// Wrap a Python `on_download_progress` argument into a core `DownloadProgressCallback`.
-///
-/// - `Some(py_callable)` → wraps it so the Python function is invoked on each chunk
-///   with `(downloaded_bytes, total_bytes)`. Exceptions are printed and swallowed.
-/// - `None` → returns `None`; core installs its own default terminal progress bar.
-///
-/// Returns `TypeError` if `py_callback` is not callable, so a non-callable argument
-/// fails fast at construction rather than per-chunk during download.
```
jona605a, line 51: Remark: `resolve_on_download_progress` lost its comment and the new `resolve_progress` is very similar. Maybe explain the difference between them, like one is for async download?

jona605a (approved):
Looks good to me. I don't have the best overview on this type of signal handling, but the implementation seems good and finely scoped.

PR #730, opened 2026-09-10
[TESTING] faster ci

duarteocarmo:
- Start Swift, Kotlin, and Python tests when their required artifacts are ready.
- Cache models and ARM64 glslc with WarpBuild; save Windows Cargo caches only on main.
- Bound model downloads and smoke tests, retain failure logs, and request whole-run cancellation on job failure.
- Preserve test coverage and model repetitions. Static checks pass; full CI will check correctness and timing.


.github/workflows/build.yml
```diff
@@ -360,6 +401,7 @@ jobs:
 
   # builds an "xcframework"
   # which is an apple thing that contains binaries for multiple systems
+      - *cancel-ci-on-failure
```
albertsgarde, line 404: This line should probably be moved up over the comments

duarteocarmo: not ready yet :)

PR #735, opened 2026-09-13
Fix FunctionGemma tool calls dropped on multi-line argument values

GuiBarradas:
## What

A FunctionGemma tool call whose argument value spans multiple lines was silently dropped: `extract_tool_calls` returned `None`, so no tool ran.

## Why

The tool-call grammar built by `to_lark` defines the value rule as `value: <escape> /[^<>{},:]+/ <escape>`. A negated character class matches a newline, so the grammar lets an argument value span several lines (a file body, a code snippet).

The extractor, however, parses the model output with two regexes that use `.`:

- `<start_function_call>\s*call:(\w+)\{(.*?)\}\s*<end_function_call>`
- `(\w+):<escape>(.*?)<escape>`

In Rust's `regex`, `.` does not match a newline without the `s` flag, so a multi-line value made the outer regex fail to match and the whole call was discarded.

FunctionGemma was the only handler with this gap: Qwen3 uses `[\s\S]*?` and Qwen3.5/3.6 uses `(?s)`.

## Fix

Add the `(?s)` flag to both regexes so `.` spans newlines. This only affects values containing a newline; every other case parses identically.

## Test

Added `test_functiongemma_extract_multiline_value`, which fails before the change (extractor returns `None`) and passes after. Verified locally with `cargo test -p nobodywho tool_calling::functiongemma::tests` (5 passed).


jona605a (approved):
Nice catch! And a clean minimal fix.
Remark: The comments read a bit dense could be shortened, but it isn't major.

PR #736, opened 2026-09-15
:construction_worker: Release tests use test applications from release tags

albertsgarde:
Previously the nightly tests on the published releases would fail any time we changed API's they depended on up until the next release. This is because we would run the updated test on the old package. This commit fixes that by making the nightly tests keep running the tests as defined at the release tags

Sadly there's no real way to test this other than to run device CI on this PR so I'll do that

Stacked on #723

Branch should be `fix-release-tests` instead of `nix-release-tests`. Oops.


albertsgarde:
Device tests failed initially but succeeded on re-run. This is not due to this PR but flakiness in the tests. Could probably be fixed by making the download process more robust since that is the part that was failing. It was caused by the phones downloading some Android stuff at the same time as the model and the Android stuff taking over.

madsmtm (approved):
I don't understand why we do these tests every night, it seems like they should only run on the release or smth.? But this is at least a good enough fix for now.

albertsgarde:
The merge conflict was a little messy so I'll run the device tests a final time before merging

albertsgarde:
Failed, but for download reasons, not because of this PR.

PR #743, opened 2026-09-16
Compatible completions (2/3): Model source

duarteocarmo:
- Store the original model load identifier in core.
- Expose read-only `source` properties across all bindings.
- Regenerate generated binding code and document the property.


duarteocarmo:
@madsmtm  I don't think we need all the docs updated etc. Just keeping this one simple.

nobodywho/core/src/llm.rs
```diff
@@ -43,12 +43,18 @@ const MAX_EMBEDDING_SEQUENCES: u32 = 256;
...
     pub(crate) projection_model: Option<ProjectionModel>,
     pub(crate) draft_model: Option<LlamaModel>,
 }
 
 impl Model {
+    /// The identifier used to load the model.
+    pub fn source(&self) -> &str {
+        &self.source
+    }
+
```
madsmtm, line 57: I considered that it might be more useful to have a `fn path()` that points to the cached model path if it's a URL, but then I thought better of it: if the user clears their cache, you don't want to try to load the now deleted file, you want to download anew.

(`fn path()` might also be useful, but let's do that sometime else, it's not necessary for this functionality)

madsmtm (approved):
Thanks!

PR #745, opened 2026-09-17
fix: `nix flake check` broken by python change

albertsgarde:
[chore: Remove deprecated](https://github.com/nobodywho-ooo/nobodywho/commit/c54f4d6ae347d9855ef4b0f9858c39f938de3bb2) broke `nix flake check` which breaks all CI on main and releases. This PR fixes it.


madsmtm (approved):
f'ing nix

nobodywho/workspace.nix
```diff
@@ -90,6 +90,11 @@ let
         # `pyo3-ffi`". We bridge the names by appending an alias to the `env` file
         # that each crate installs and its dependents source.
         pyo3-ffi = attrs: {
+          # Stands in for the removed `pyo3/extension-module` cargo feature, which
```
madsmtm, line 93: @duarteocarmo fixed this in https://github.com/nobodywho-ooo/nobodywho/pull/711 with:
```
nativeBuildInputs = (attrs.nativeBuildInputs or [ ]) ++ [ pkgs.python3 ];
```

Would that be better or worse?

albertsgarde: According to Claude, what this PR does leads to behaviour identical to what we had before. It's what Pyo3 themselves suggest: https://pyo3.rs/v0.29.2/building-and-distribution.html#the-pyo3_build_extension_module-environment-variable

As I understand it the other fix kinda locks in a specific libpython dependency which might then conflict with other things.

PR #748, opened 2026-09-17
Add Python logging control

duarteocarmo:
Multimodal logs were not supressed in Python - this fixes that.


madsmtm (changes requested):
I've already fixed this in https://github.com/utilityai/llama-cpp-rs/pull/1104, it'd be simpler to update `llama-cpp-2`?

duarteocarmo:
@madsmtm  ah ok - didn't know! IF we are addressing, great.

madsmtm:
I've put actually updating `llama-cpp-2` as a TODO on my Notion

PR #752, opened 2026-09-18
Update llama-cpp-2 to 0.1.156

madsmtm:
In particular, this contains the fix for https://github.com/nobodywho-ooo/nobodywho/pull/748.

This also bumps llama.cpp from a commit on 2026-07-30 to one on 2026-08-13.


duarteocarmo (approved):

duarteocarmo (approved):
Just make sure we dont have to update penalty docs anywhere else for the bindings

